PR #2, opened 2024-05-13
Introduce an interface to translate hostmetrics and add tests

lahsivjar:
Hostmetrics remappers remap metrics produced by [hostmetricsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver) to [Elastic system metrics](https://www.elastic.co/guide/en/beats/metricbeat/current/exported-fields-system.html). This allows powering Elastic's curated UIs based on OTel metrics.

The code from the remappers is adopted from the [PoC done here](https://github.com/tommyers-elastic/opentelemetry-collector-contrib/tree/routing-processor/processor/elasticprocessor/internal/hostmetrics).


remappers/const.go
```diff
@@ -0,0 +1,27 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package remappers
+
+const (
+	// OTelTranslatedLabel is added as an attribute to each datapoint that
+	// is translated from otel.
+	OTelTranslatedLabel = "otel_translated"
```
lahsivjar, line 23: [**For reviewers**] This label is added to all data points that are created using the remapping logic. Suggestions for the label name is welcome if the current one doesn't look good.

In a followup PR, I plan to add the same label to the source metric's scope attributes and filter the attribute to avoid double translation.

remappers/hostmetrics/config.go
```diff
@@ -0,0 +1,41 @@
...
+func newConfig(opts ...Option) (cfg config) {
+	for _, opt := range opts {
+		cfg = opt(cfg)
+	}
+	return cfg
+}
+
+// WithSystemIntegrationDataset sets the dataset of the remapped metrics as
+// as per the system integration. Example: system.cpu, system.memory, etc.
+func WithSystemIntegrationDataset(b bool) Option {
```
lahsivjar, line 36: [**For reviewers**] This option allows setting the dataset based as required by the system integration. With this option enabled, the metrics should end up in system integration datastreams - note that I have not tested this option.

remappers/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,107 @@
...
+	case "cpu":
+		err = remapCPUMetrics(src.Metrics(), out, dataset)
+	case "memory":
+		err = remapMemoryMetrics(src.Metrics(), out, dataset)
+	case "load":
+		err = remapLoadMetrics(src.Metrics(), out, dataset)
+	case "process":
+		err = remapProcessMetrics(src.Metrics(), out, dataset)
+	default:
+		r.logger.Warn("no remapper found for scope", zap.String("scope", scope.Name()))
```
axw, line 91: Should we add no-op cases for the scopes that we know don't (yet) have translations? Otherwise there will be a lot of warnings that we can't do anything about...?

remappers/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,107 @@
...
+		cfg:    newConfig(opts...),
+		logger: logger,
+	}
+}
+
+// Remap remaps an OTel ScopeMetrics to a list of OTel metrics such that the
+// remapped metrics could be trivially converted into Elastic system metrics.
+// The current remapping logic assumes that each Metric in the ScopeMetric
+// will have datapoints for a single timestamp only.
+func (r *Remapper) Remap(src pmetric.ScopeMetrics, out pmetric.MetricSlice) {
```
axw, line 62: What is the reason for not implementing an opentelemetry-collector processor directly here?

axw (commented):
Just a high level review - looks good overall, main questions are whether we should be resilient to changes in the set of host metrics scraped, and whether this should be implementing the processor API directly rather than creating a new "remapper" abstraction.

remappers/hostmetrics/process.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	addMetrics(out, dataset,
+		metric{
+			dataType:  pmetric.MetricTypeSum,
+			name:      "system.process.num_threads",
+			timestamp: timestamp,
+			intValue:  &threads,
+		},
+	)
```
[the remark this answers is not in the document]
lahsivjar: > What if there are no process.threads metric data points?

Ah, `process.threads` is an optional metric ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processscraper/documentation.md#processthreads)). From the PoC I fixed the code to assume the optional metrics might be missing but missed this, I will add a check here.

> Should we handle the case where a user has filtered some `hostmetrics out before sending on to Elastic?

The remapping codes assume that each `ScopeMetric` will have all the non-default metrics from the scraper and with a single timestamp. Both these assumptions will be impacted if we were to assume a batching/filtering processor in-between, let me see if I can perform some checks. Initially I was thinking to transform each datapoint in a more generic way but our requirement to calculate normalized metrics made it a lot difficult to do that.

remappers/hostmetrics/process.go
```diff
@@ -0,0 +1,48 @@
...
+func remapProcessMetrics(src, out pmetric.MetricSlice, dataset string) error {
+	var timestamp pcommon.Timestamp
+	var threads int64
+
+	for i := 0; i < src.Len(); i++ {
+		metric := src.At(i)
+		if metric.Name() == "process.threads" {
+			dp := metric.Sum().DataPoints().At(0)
+			timestamp = dp.Timestamp()
+			threads = dp.IntValue()
```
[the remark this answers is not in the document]
lahsivjar: Hmm, for this specific case, we can break it but we would probably need to handle other metrics in the future. For example: `process.cpu.time`

remappers/hostmetrics/hostmetrics.go
lahsivjar: Good point. I will make this a no-op.

remappers/hostmetrics/hostmetrics.go
lahsivjar: Do you mean the full processor interface ([ref](https://github.com/open-telemetry/opentelemetry-collector/blob/671dcbcac3073c8001836e4b90354c345fce0d69/processor/processor.go#L28-L31))? -- I didn't do this because we didn't need to. The `component.Component` felt unnecessary complexity for our use-case.

The consumer interface ([ref](https://github.com/open-telemetry/opentelemetry-collector/blob/671dcbcac3073c8001836e4b90354c345fce0d69/consumer/metrics.go#L14-L18)) was a better alternative but I decided against it because it would be too restrictive. My goal was to have an interface that can be used in a multiple ways and easily adapted to implement a consumer interface if needed - the reason why I separated the output `pmetric.MetricSlice` from the source's `pmetric.ScopeMetrics`. Let me know if it doesn't make sense, I can adopt the consumer interface.

lahsivjar: > It's not a big deal, and we can refactor as needed. I was wondering if there was something fundamentally preventing us - sounds like no.

Yeah, no blockers in doing that.

lahsivjar:
Putting this PR on draft after talking with @ishleenk17. Once https://github.com/elastic/opentelemetry-lib/pull/4 is merged I will rebase the PR and update the code.

lahsivjar:
Apologies for the force push, I have updated the PR after the original code for the remappers was merged. It is ready for review now.

remappers/hostmetrics/process.go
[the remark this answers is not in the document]
lahsivjar: There are other metrics added to the `process` scraper now so this doesn't hold true anymore.

remappers/hostmetrics/process.go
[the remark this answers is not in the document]
lahsivjar: > Should we handle the case where a user has filtered some `hostmetrics out before sending on to Elastic?

Currently, the remapper will just produce `0` @axw by handling do you mean not producing metric if it is not initialized (missing) right?

remappers/hostmetrics/cpu.go
```diff
@@ -61,23 +85,8 @@ func addCPUMetrics(metrics pmetric.MetricSlice, resource pcommon.Resource, datas
...
-
-	addMetrics(metrics, resource, dataset,
-		metric{
-			dataType:  pmetric.MetricTypeSum,
-			name:      "system.cpu.cores",
-			timestamp: timestamp,
-			intValue:  &numCores,
-		},
+	// Add all metrics that are independent of cpu logical count.
+	addMetrics(out, dataset, emptyMutator,
```
ishleenk17, line 89: We are putting out the mtrics in different metric slices here?

remappers/hostmetrics/cpu.go
```diff
@@ -132,6 +141,38 @@ func addCPUMetrics(metrics pmetric.MetricSlice, resource pcommon.Resource, datas
 			timestamp:   timestamp,
 			doubleValue: &softirqPercent,
 		},
+	)
+
+	// TODO (lahsivjar): Remove the dependency on the system.cpu.logical.count
```
ishleenk17, line 146: Are we gaining anything if we do this ?
Since system.cpu.utilization is also an optional metric ?

remappers/hostmetrics/cpu.go
```diff
@@ -132,6 +141,38 @@ func addCPUMetrics(metrics pmetric.MetricSlice, resource pcommon.Resource, datas
...
+	addMetrics(out, dataset, emptyMutator,
+		metric{
+			dataType:  pmetric.MetricTypeSum,
+			name:      "system.cpu.cores",
+			timestamp: timestamp,
+			intValue:  &numCores,
+		},
+		metric{
+			dataType:  pmetric.MetricTypeSum,
+			name:      "system.load.cores",
```
ishleenk17, line 172: This would be used for ?

remappers/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,114 @@
...
+	case "memory":
+		err = remapMemoryMetrics(src.Metrics(), out, resource, dataset)
+	case "load":
+		err = remapLoadMetrics(src.Metrics(), out, resource, dataset)
+	case "process":
+		err = remapProcessMetrics(src.Metrics(), out, resource, dataset)
+	case "processes":
+		err = remapProcessesMetrics(src.Metrics(), out, resource, dataset)
+	case "network":
+		err = remapProcessesMetrics(src.Metrics(), out, resource, dataset)
```
ishleenk17, line 98: Please call the NetowrkMetrics here

remappers/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,114 @@
...
+// will have datapoints for a single timestamp only.
+func (r *Remapper) Remap(
+	src pmetric.ScopeMetrics,
+	out pmetric.MetricSlice,
+	resource pcommon.Resource,
+) {
+	if !r.Valid(src) {
+		return
+	}
+
```
ishleenk17, line 71: The metricset.period calculation from the migrated code has been left out here. Was that intentional ?

remappers/hostmetrics/metric.go
```diff
@@ -0,0 +1,75 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
```
ishleenk17, line 8: Any reason why we can't use the existing metric.go file ?
The code looks very similar with minor changes. 
Also we would like to have the attributes in the structure as in the network, disk, process scraper etc. we would be using them. So, I think its prudent to edit in the existing metric.go file

remappers/hostmetrics/metric.go
```diff
@@ -0,0 +1,75 @@
...
+		} else if metric.doubleValue != nil {
+			dp.SetDoubleValue(*metric.doubleValue)
+		}
+
+		dp.SetTimestamp(metric.timestamp)
+		if metric.startTimestamp != 0 {
+			dp.SetStartTimestamp(metric.startTimestamp)
+		}
+
+		dp.Attributes().PutBool(common.OTelTranslatedLabel, true)
```
ishleenk17, line 68: Should we add it to one of the existing ecs fields ? Like you can see in thee migrated metric.go file.
Initially we had thought to use event.provider, but we can discuss.

remappers/hostmetrics/network.go
```diff
@@ -0,0 +1,88 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
```
ishleenk17, line 5: Any reason to not use the migrated network.go file ?

ishleenk17 (commented):
Thanks Vishal!
I've left my initial comments on the PR. I'll be doing a more thorough review soon.

Also, let's change the title of the PR to something more appropriate.

remappers/hostmetrics/cpu.go
lahsivjar: Yes, you can choose to use the same src by passing in the `MetricSlice` created from the `ScopeMetric`

remappers/hostmetrics/cpu.go
lahsivjar: The code would have more complexity and branches when we depend on 2 metrics vs 1 metric.

remappers/hostmetrics/hostmetrics.go
lahsivjar: Ah, I am missing tests for the `network` 🤦

remappers/hostmetrics/cpu.go
lahsivjar: It is used in the load dasbhoard as a denominator in one of the panels.

remappers/hostmetrics/hostmetrics.go
lahsivjar: Not intentional actually. Can you elaborate on how the `metricset.period` is used?

remappers/hostmetrics/metric.go
lahsivjar: > Any reason why we can't use the existing metric.go file ?

It is how git shows the diff when I moved the file, I have not actually removed it.

> Also we would like to have the attributes in the structure as in the network, disk, process scraper etc. we would be using them.

How exactly would the `addMetric` use them, is the `mutator` logic not good enough? The [current code](https://github.com/elastic/opentelemetry-lib/blob/fdb46e794d6b6b95b2edf5f1fbe43e423e4eed2e/hostmetrics/metric.go#L53-L57) in the metric uses conditionals to add process-specific attributes which leaks dataset specific logic in the more generic `addMetric` function. The PR introduces a `mutator` func which you can pass in to add any scraper-specific attributes/logic.

remappers/hostmetrics/metric.go
lahsivjar: Using existing ecs fields makes sense, I initially based my code from https://github.com/ishleenk17/opentelemetry-collector-contrib/tree/ishleen_processor and missed the logic when rebasing. Let me move to `event.provider`. For the value, should we use `otel/hostmetrics`?

remappers/hostmetrics/network.go
lahsivjar: Same as above, I moved the file but git shows it as delete + add.

remappers/hostmetrics/cpu.go
```diff
@@ -1,27 +1,51 @@
...
 
 import (
+	"errors"
+
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 )
 
-func addCPUMetrics(metrics pmetric.MetricSlice, resource pcommon.Resource, dataset string) error {
+func remapCPUMetrics(
```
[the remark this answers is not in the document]
lahsivjar: Hmm, we are remapping + adding so both miss one part of the equation. I don't have a strong preference either way, if you feel strongly for `Add` I will make the changes OR maybe we can come up with a better name 🤔

remappers/hostmetrics/metric.go
lahsivjar: Update it to use `event.provider: hostmetrics`.

remappers/hostmetrics/cpu.go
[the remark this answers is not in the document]
lahsivjar: Actually, the interface itself just remaps (or maybe translate?) the OTel metric and returns it as `out`. What you do with it is left to the caller so I don't think `Add` is correct here.

remappers/hostmetrics/hostmetrics.go
lahsivjar: I perused the actual usage [here](https://github.com/tommyers-elastic/opentelemetry-collector-contrib/blob/fdd9e5b13bd0d8b31bab48dd394755d8aad58683/processor/elasticprocessor/processor.go#L36-L40) and I think this piece of code would better live outside the remappers so that the interfaces don't get bloated. Let me know what you think.

remappers/hostmetrics/cpu.go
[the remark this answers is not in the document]
ishleenk17: hmm, I have mixed thoughts on this.
@tommyers-elastic WDYT ?

remappers/hostmetrics/process.go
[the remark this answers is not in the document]
axw: Right, I think if it's not in the input it shouldn't be in the output.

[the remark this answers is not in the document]
lahsivjar: Got it, now that the remappers are already in the main branch I have opened an issue for this: https://github.com/elastic/opentelemetry-lib/issues/6

remappers/hostmetrics/cpu.go
lahsivjar: I have removed the TODO and [created an issue](https://github.com/elastic/opentelemetry-lib/issues/7) to discuss this so that the current PR can be reviewed quickly. Resolving this.

remappers/hostmetrics/hostmetrics.go
tommyers-elastic: i added metricset.period because it's used in some of the UIs as a scale factor. i agree it's kinda clunky here - ideally you'd be able to read it from the receiver config (the scrape period) directly, but there's no way i could find to it. 

we could add it in a downstream component, but i figured as it was a direct requirement of some of the elastic host UIs, it was ok to stick it here for now.

tommyers-elastic (approved):
looking good

remappers/hostmetrics/process.go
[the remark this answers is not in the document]
ishleenk17: > Right, I think if it's not in the input it shouldn't be in the output.

Makes sense @axw . Thanks for the issue Vishal.

remappers/hostmetrics/hostmetrics.go
lahsivjar: Hmm, I think the issue with doing it in library (or even in APM/MIS) would be that we would need to maintain a cache for all the hosts we see and calculate the `metricset.period` for each of it. @axw WDYT? Do you see a better way here?

> but i figured as it was a direct requirement of some of the elastic host UIs, it was ok to stick it here for now

Do you have a list of UIs that break due to this? I checked the main inventory UI and that seems to work fine.

remappers/hostmetrics/cpu.go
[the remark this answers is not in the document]
tommyers-elastic: as far as i remember the operation here is purely addative - we add new data points into existin scopemetrics, right? this was the reason for the original choice of wording. let's not loose sleep over this, it doesn't really matter.

remappers/hostmetrics/hostmetrics.go
tommyers-elastic: it was the "hosts" page and the "host detail" page. 

in the poc implementation we just stored a single value for the whole processor. this worked fine - it assumes that we only have one pmetrics struct per scrape (across all the datasets), which is currently true. could you think of a configuration which would break it?

remappers/hostmetrics/cpu.go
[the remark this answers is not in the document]
lahsivjar: > we add new data points into existin scopemetrics, right? this was the reason for the original choice of wording

The current interface does it in a different way. The interface accepts source as `pmetric.ScopeMetrics` and output as `pmetric.MetricSlice` (you can chose to use the same metric slice by just passing in `ScopeMetrics#Metrics()` if you wan to put in existing scope metrics. So, the operations performed are read the metrics from the source -> remap it -> write it to output.

remappers/hostmetrics/hostmetrics.go
lahsivjar: > It doesn't look like we're dealing with cumulative sums in this code.

We are dealing with cumulative sums (in network and process remappers) :(

remappers/hostmetrics/cpu.go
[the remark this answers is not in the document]
lahsivjar: However, I agree we don't need to spend to much time on renaming. @ishleenk17  let me know if `Remap` feels like a deal-breaker to you, I will rename it to `Add`.

remappers/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,115 @@
...
+		logger: logger,
+	}
+}
+
+// Remap remaps an OTel ScopeMetrics to a list of OTel metrics such that the
+// remapped metrics could be trivially converted into Elastic system metrics.
+// The current remapping logic assumes that each Metric in the ScopeMetric
+// will have datapoints for a single timestamp only. The remapped metrics are
+// added to the output `MetricSlice`.
+func (r *Remapper) Remap(
```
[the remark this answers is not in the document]
lahsivjar: Done!

remappers/hostmetrics/hostmetrics.go
lahsivjar: > in the poc implementation we just stored a single value for the whole processor. this worked fine - it assumes that we only have one pmetrics struct per scrape (across all the datasets), which is currently true. could you think of a configuration which would break it?

@tommyers-elastic But in APM/MIS we would get data from more than one collector instances, so we would need to calculate the `metricset.period` for all such instances even if it is 1 value per instance. Did I answer your question?

remappers/hostmetrics/cpu.go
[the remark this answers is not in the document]
ishleenk17: Lets go ahead with Remap!

remappers/hostmetrics/hostmetrics.go
ishleenk17: @lahsivjar : Lets open a ticket to discuss about metricset.period, as this should not block the PR

ishleenk17 (approved):
Looks good!

remappers/hostmetrics/hostmetrics.go
axw: > We are dealing with cumulative sums (in network and process remappers) :(

Right, that's a pain. One option, which is a little brittle but probably not worse than other options, would be to extract the value from any of the non-cumulative sum metrics. All of the host metrics are scraped on the same interval IIANM.

axw (approved):

remappers/hostmetrics/hostmetrics.go
ishleenk17: @axw : I have created a separate [issue](https://github.com/elastic/opentelemetry-lib/issues/8) for this discussion to unblock this PR. 
Lets take it forward there.

ishleenk17: > @lahsivjar : Lets open a ticket to discuss about metricset.period, as this should not block the PR

The ticket for [metricset.period](https://github.com/elastic/opentelemetry-lib/issues/9).

remappers/hostmetrics/metric.go
```diff
@@ -0,0 +1,75 @@
...
+	metrics ...metric,
+) {
+	ms.EnsureCapacity(ms.Len() + len(metrics))
+
+	for _, metric := range metrics {
+		m := ms.AppendEmpty()
+		m.SetName(metric.name)
+
+		var dp pmetric.NumberDataPoint
+		switch metric.dataType {
```
shmsr, line 50: shouldn't there be a default case to skip the metric if any other dataType comes other than MetricTypeGauge/ MetricTypeSum?

remappers/hostmetrics/network.go
```diff
@@ -0,0 +1,88 @@
...
+				timestamp := dp.Timestamp()
+				value := dp.IntValue()
+
+				switch direction.Str() {
+				case "receive":
+					addDeviceMetric(out, timestamp, dataset, name, device, "in", value)
+				case "transmit":
+					addDeviceMetric(out, timestamp, dataset, name, device, "out", value)
+				}
+			}
```
shmsr, line 56:
```suggestion
			dp := dataPoints.At(j)
			device, ok := dp.Attributes().Get("device")
			if !ok {
				continue
			}

			direction, ok := dp.Attributes().Get("direction")
			if !ok {
				continue
			}

			name := metric.Name()
			timestamp := dp.Timestamp()
			value := dp.IntValue()

			switch direction.Str() {
			case "receive":
				addDeviceMetric(out, timestamp, dataset, name, device.Str(), "in", value)
			case "transmit":
				addDeviceMetric(out, timestamp, dataset, name, device.Str(), "out", value)
			}
```

Above logic can be simplified to this?

remappers/hostmetrics/network.go
```diff
@@ -0,0 +1,88 @@
...
+	value int64,
+) {
+	metricsToAdd := map[string]string{
+		"system.network.io":      "system.network.%s.bytes",
+		"system.network.packets": "system.network.%s.packets",
+		"system.network.dropped": "system.network.%s.dropped",
+		"system.network.errors":  "system.network.%s.errors",
+	}
+
+	if metricNetworkES, ok := metricsToAdd[name]; ok {
```
shmsr, line 76: [invert if](https://github.com/golang/vscode-go/issues/2557#issue-1474966959) for better readability?

```go
metricNetworkES, ok := metricsToAdd[name]
if !ok {
	return
}

addMetrics(out, dataset, func(dp pmetric.NumberDataPoint) {
	dp.Attributes().PutStr("system.network.name", device)
}, metric{
	dataType:  pmetric.MetricTypeSum,
	name:      fmt.Sprintf(metricNetworkES, direction),
	timestamp: timestamp,
	intValue:  &value,
})
```

shmsr (commented):
Left some minor comments.

remappers/hostmetrics/network.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+)
+
+func remapNetworkMetrics(
+	src, out pmetric.MetricSlice,
+	resource pcommon.Resource,
```
shmsr, line 29:
```suggestion
	_ pcommon.Resource,
```

remappers/hostmetrics/hostmetrics.go
```diff
@@ -0,0 +1,116 @@
...
+		err = remapNetworkMetrics(src.Metrics(), out, resource, dataset)
+	}
+
+	if err != nil {
+		r.logger.Warn(
+			"failed to remap OTel hostmetrics",
+			zap.String("scope", scope.Name()),
+			zap.Error(err),
+		)
+	}
```
shmsr, line 109: This can be simplified to not call `remapNetworkMetrics(src.Metrics(), out, resource, dataset)` repeat 6 times.

```go
remapFunc, ok := remapFuncs[scraper]
if !ok {
// If no remapping function is defined for the scraper, do something?
}

err := remapFunc(src.Metrics(), out, resource, dataset)
if err != nil {
r.logger.Warn(
	"failed to remap OTel hostmetrics",
	zap.String("scope", scope.Name()),
	zap.Error(err),
)
}
```

And the following in global scope:
```go
type remapFunc func(metrics pmetric.MetricSlice, out pmetric.MetricSlice, resource pcommon.Resource, dataset string) error

var remapFuncs = map[string]remapFunc{
	"cpu":       remapCPUMetrics,
	"memory":    remapMemoryMetrics,
	"load":      remapLoadMetrics,
	"process":   remapProcessMetrics,
	"processes": remapProcessesMetrics,
	"network":   remapNetworkMetrics,
}
```

lahsivjar: Hmm, it is a repeat either way and I personally see them as similar reading complexity. In fact, using a remap func will have an additional lookup to find which code is being called but I don't have a strong opinion here. Updated the code.

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	go test -v -race ./...
+
+fmt:
+	go run golang.org/x/tools/cmd/goimports@v0.18.0 -w .
+
+gomodtidy:
+	go mod tidy -v
+
+fieldalignment:
+	go run golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@v0.18.0 -test=false $(shell go list ./... | grep -v test)
```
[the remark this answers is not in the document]
lahsivjar: Ah, good catch. Updated it to use `v0.21.0`. BTW, `fieldalignment` is not getting used right now by the CI so I will need to create a followup PR to add these to CI.

remappers/hostmetrics/metric.go
lahsivjar: The code shows as diff here but it is same as what we have in `main`. IMO, the best approach would be to throw an error (or even panic here) since skipping would mean we have appended an empty metric to the slice with a name but didn't handle it -- this would only occur when there is a bug in the library and the tests don't catch it for both the cases. I will leave it as is for now and will open another PR to fix it.

Makefile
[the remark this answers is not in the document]
lahsivjar: > Can we use @latest?

Not a fan of using `latest` so updated to `v0.21.0`, let me know if you have a strong inclination towards `latest`.

remappers/hostmetrics/metric.go
ishleenk17: We will take it as part of next refactoring Subham.

Makefile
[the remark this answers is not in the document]
shmsr: No 0.21.0 is fine. I also do not have strong inclination towards `latest`.

PR #10, opened 2024-05-17
Make metrics reusable by other remappers

lahsivjar:
`AddMetrics` function would be required by other mappers, migrate it to an internal package to allow this.


remappers/internal/metric.go
```diff
@@ -23,46 +23,49 @@ import (
...
 
-		dp.SetTimestamp(metric.timestamp)
-		if metric.startTimestamp != 0 {
-			dp.SetStartTimestamp(metric.startTimestamp)
+		dp.SetTimestamp(metric.Timestamp)
+		if metric.StartTimestamp != 0 {
+			dp.SetStartTimestamp(metric.StartTimestamp)
 		}
 
 		dp.Attributes().PutStr("event.provider", "hostmetrics")
```
ishleenk17, line 71: Lets add the change to event.module in this PR itself. 
```suggestion
		dp.Attributes().PutStr("event.module", "elastic/opentelemetry-lib")
```

ishleenk17:
Changes look good. Left one comment, after addressing that we can merge it.
@lahsivjar : I hope we have tested it as well after reorganizing the structure.

remappers/internal/metric.go
lahsivjar: I prefer to create a separate PR for that.

ishleenk17 (approved):
Looks good!

PR #18, opened 2024-05-31
Add event.module attribute to all remapped metrics

lahsivjar:
Based on discussions with the team we will be adopting the [event.module](https://www.elastic.co/guide/en/ecs/current/ecs-event.html#field-event-module) ECS attribute instead of `event.provider` to taint our remapped metrics.

### DIFF
--- a/remappers/internal/metric.go
+++ b/remappers/internal/metric.go
@@ -68,7 +68,7 @@ func AddMetrics(
 			dp.SetStartTimestamp(metric.StartTimestamp)
 		}
 
-		dp.Attributes().PutStr("event.provider", "hostmetrics")
+		dp.Attributes().PutStr("event.module", common.RemapperEventModule)
 		if dataset != "" {
 			dp.Attributes().PutStr(common.DatastreamDatasetLabel, dataset)
 		}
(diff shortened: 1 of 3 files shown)


remappers/internal/metric.go
```diff
@@ -68,7 +68,7 @@ func AddMetrics(
 			dp.SetStartTimestamp(metric.StartTimestamp)
 		}
 
-		dp.Attributes().PutStr("event.provider", "hostmetrics")
+		dp.Attributes().PutStr("event.module", common.RemapperEventModule)
```
ishleenk17, line 71: The only issue that might happen is if there is a panel/visualization in UI's using event.module for representation since this value is "system" in the System Integration.

lahsivjar: Hmm, can we confirm this in some way? If not then what do you suggest we should use?

ishleenk17: I don't see it being used in dashboards, the issue would be for custom system dashboards, if someone has built a dashboard using this field name. 
In that case, having a new field might make sense. 
@tommyers-elastic : WDYT ?

ishleenk17: Lets go ahead with event.module

ishleenk17 (approved):
Looks good!

PR #20, opened 2024-06-04
Adding k8s remapper for supporting Kibana Inventory "Kubernetes Pod" page 

gizas:
This is the main PR needed for https://github.com/elastic/opentelemetry-dev/issues/180

It adds the remapping functionality for kubeletstats and k8scluster receivers


gizas:
cc @AlexanderWert , @ChrsMark , @tetianakravchenko 

This is the relevant PR for the remapping of k8s otel metrics in opentelemetry-lib. Lets sync on how to test it

ChrsMark:
Nice @gizas !

> This is the relevant PR for the remapping of k8s otel metrics in opentelemetry-lib. Lets sync on how to test it

I guess we would need to import the `opentelemetry-lib` in the Elastic Specific processor in order to test this e2e?
I assume this Elastic Processor would need to be part of the https://github.com/elastic/opentelemetry-collector-components but I'm not 100% sure if and when this will happen.

@AlexanderWert should we prioritize the Elastic processor so as different teams/contributors can have this as base for the specific implementations that would be on top?

remappers/k8smetrics/k8smetrics.go
```diff
@@ -0,0 +1,88 @@
...
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.uber.org/zap"
+)
+
+type remapFunc func(metrics pmetric.MetricSlice, out pmetric.MetricSlice, resource pcommon.Resource) error
+
+// Remapper maps the OTel hostmetrics to Elastic system metrics. These remapped
+// metrics power the curated Kibana dashboards. Each datapoint translated using
+// the remapper has the `event.processor` attribute set to `kubernetes`.
+type Remapper struct {
```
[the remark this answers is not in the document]
gizas: it inlcudes the `cfg    config` variable that is specific to each package, eg.

 type config struct {
	KubernetesIntegrationDataset bool
}

I would say to leave it for now as it would require other changes. WDYT?

remappers/internal/metric_k8s.go
```diff
@@ -0,0 +1,60 @@
+package internal
```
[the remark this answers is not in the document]
gizas: I removed metrics_k8s.go file and moved all inside metrics.go

AlexanderWert:
> @AlexanderWert should we prioritize the Elastic processor so as different teams/contributors can have this as base for the specific implementations that would be on top?

IIUC, @ishleenk17 is already working on that

ishleenk17:
> > @AlexanderWert should we prioritize the Elastic processor so as different teams/contributors can have this as base for the specific implementations that would be on top?
> 
> IIUC, @ishleenk17 is already working on that

The code is ready in the private branch and I have pointed @gizas to use that for testing.
In the meantime, the [PR](https://github.com/elastic/opentelemetry-collector-components/pull/7) is being reviewed for merging this code into the [components](https://github.com/elastic/opentelemetry-collector-components) repo.

tetianakravchenko:
@gizas FYI: I've tested this PR - https://github.com/tetianakravchenko/opentelemetry-collector-contrib/commit/808f7f179547910c5897e23f3c9b5346f353ebf3 (changes are based on the mentioned above private branch - https://github.com/tommyers-elastic/opentelemetry-collector-contrib/tree/routing-processor and use the reference to the local [`opentelemetry-lib`](https://github.com/tetianakravchenko/opentelemetry-collector-contrib/commit/808f7f179547910c5897e23f3c9b5346f353ebf3#diff-54b883f5082927195f91c8ecc18bbd28edd3eb7f017009f79bce8c4712bd9b4eR43))

all works:
Screenshot 2024-06-10 at 17 58 46
(k8s integration assets and the [ingest pipeline](https://github.com/gizas/opentelemetry-collector-contrib/blob/k8smetrics/processor/elasticprocessor/templates.json) were installed to make it work)

remappers/internal/metric.go
```diff
@@ -76,3 +78,55 @@ func AddMetrics(
 		mutator(dp)
 	}
 }
+
+func Addk8sMetrics(ms pmetric.MetricSlice, mutator func(dp pmetric.NumberDataPoint), metrics ...Metric) {
```
lahsivjar, line 82: I think we can use the `AddMetrics` here. The kubernetes specific logic should be added in the `mutator` and be kept isolated in the `kubernetesmetrics` pacakge.

remappers/internal/metric.go
```diff
@@ -76,3 +78,55 @@ func AddMetrics(
...
+		}
+
+		// Calculate datastream attribute as an attribute to each datapoint
+		dataset := AddDatastream(metric.Name)
+		if dataset == "kubernetes.node" {
+			dataset = "kubernetes.pod"
+		}
+		dp.Attributes().PutStr("event.module", common.RemapperEventModule)
+		dp.Attributes().PutStr("service.type", "kubernetes")
+		dp.Attributes().PutStr(common.DatastreamDatasetLabel, dataset)
```
lahsivjar, line 115: For APM-Server logic we should not be setting the dataset since we want the dataset to be set via APM-Server so as to remove the dependency on any external integration for the APM-Server path. The config option `WIthSystemIntegrationDataset` for for this purpose -- when it is set to false then we won't need to deduce the dataset in the remapper.

remappers/internal/metric.go
```diff
@@ -76,3 +78,55 @@ func AddMetrics(
...
+			dp.SetStartTimestamp(metric.StartTimestamp)
+		}
+
+		// Calculate datastream attribute as an attribute to each datapoint
+		dataset := AddDatastream(metric.Name)
+		if dataset == "kubernetes.node" {
+			dataset = "kubernetes.pod"
+		}
+		dp.Attributes().PutStr("event.module", common.RemapperEventModule)
+		dp.Attributes().PutStr("service.type", "kubernetes")
```
lahsivjar, line 114: This is specific to kubernetes and should be moved to the mutator and the function `Addk8sMetrcs` should be same as the `AddMetrics`.

lahsivjar (commented):
I don't think we need to have a new function name `AddK8sMetrics` and this will not work with the APM-Server ingestion path due to the dataset being explicitly set.

remappers/internal/metric.go
lahsivjar: I see that the kubernetes specific logic here is setting dataset which is dependent on the metric name. This would not be possible to set using the current design of the mutator which only takes a datapoint. There are 2 options we could consider:

1. Update the `Metric` model to have a `Dataset` field and change the signature of the `AddMetric` to `func(ms pmetric.MetricSlice, mutator func(dp pmetric.NumberDataPoint), metrics ...Metric)`. Then each metric can have its own dataset. Note that we would need to consider the config option and set empty dataset if the flag `WithKubernetesIntegrationDataset` is not set.
2. Update the design of the mutator to take to `func(m pmetric.Metric)` and use the mutator to do the k8s specific dataset logic.

I would personally prefer `1` because it is cleaner IMO and dataset is specific to each metric by our design but I will leave it up to you to decide which is better.

remappers/internal/metric.go
```diff
@@ -70,6 +70,9 @@ func AddMetrics(
 
 		dp.Attributes().PutStr("event.module", common.RemapperEventModule)
 		if dataset != "" {
+			if dataset == "kubernetes.pod" {
+				dp.Attributes().PutStr("service.type", "kubernetes")
+			}
```
lahsivjar, line 75: We can move this logic to the `remappers/kubernetesmetrics/kubelet.go` by using a custom mutator instead of `remappers.EmptyMutator`. The mutator will be something like:

```go
func(dp pmetric.NumberDataPoint) {
	dp.Attributes().PutStr("service.type", "kubernetes")
}
```

gizas: @lahsivjar thank you done as well. We are currently testing the latest changes and let u know if sth is broken or not

lahsivjar: Great! Thanks for the changes, one thing remaining is that we are currently not obeying the config `WithKubernetesIntegrationDataset`. I think we can follow an approach similar to https://github.com/elastic/opentelemetry-lib/blob/76a6d08387f09b1372328f5439fdd43a0bfd2670/remappers/hostmetrics/hostmetrics.go#L89-L97 for deducing the dataset based on the config option.

We require to not set dataset for the OTel-Collector -> APM-Server|MIS -> ES path where the dataset would be set by the APM/MIS libraries.

gizas: I dont think is feasible in k8s case as in hostmetrics is defined based on the scraper that is provided. And it is 1-1 relationship between scraper and dataset
In k8s case this is not valid. I mean we have 2 scrapers (kubeletstatsreceiver and k8sclusterreceiver) but these can save even to same dataset. How about leaving it as it is?

lahsivjar: > And it is 1-1 relationship between scraper and dataset
> In k8s case this is not valid

Sorry if I am missing something obvious here but to me it looks like the latest code uses the dataset `kubernetes.pod` for `addKubeletMetrics` and `kubernetes.node` for `addClusterMetrics`. If this is true then we can deduce the data set and pass it directly to the remappers.

>  How about leaving it as it is?

I am okay with this too though we would still need to handle `WithKubernetesIntegrationDataset` option where the dataset should be set only if the option is set to `true`, right?

remappers/kubernetesmetrics/k8smetrics_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/elastic/opentelemetry-lib/remappers/common"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/tj/assert"
```
dmathieu, line 28: This seems like the wrong import? Other packages are using stretchr.
If the intent is to have requires, there's the [require package](https://pkg.go.dev/github.com/stretchr/testify/require) that would avoid adding another dependency to the repository.

gizas: @dmathieu nw we crossed messages here. It is fixed now with latest commit

remappers/internal/metric.go
gizas: @lahsivjar this is accurate at the moment:
addKubeletMetrics -> kubernetes.pod
addClusterMetrics -> kubernetes.node

And actually the clustermetrics are not even used in the UI. But we decided to inlcude them for now as there are some references of those inside Kibana code. 

But in general the kubelet can route to many datastreams. Although for now this is not the case I agree. I just dont want to introduce something now that we might change in the future. Let me also discuss this internally and can come back with you.

In general we have built locally the changes and this code works

lahsivjar: Got it. Makes sense to not use the same approach as we do for the hostmetrics.

We still need to obey the config option `WithKubernetesIntegrationDataset`, so maybe we can pass the config option to the `addKubeletMetrics` and `addClusterMetrics` where the dataset can be set to empty if the option is `false`.

lahsivjar:
FWIW, to fix the lint we would need to update the notices which could be done by either running `make all` OR `make update-licenses`.

remappers/kubernetesmetrics/clustermetrics.go
```diff
@@ -0,0 +1,52 @@
...
+
+func addClusterMetrics(
+	src, out pmetric.MetricSlice,
+	_ pcommon.Resource,
+) error {
+	var timestamp pcommon.Timestamp
+	var node_allocatable_memory, node_allocatable_cpu int64
+
+	// iterate all metrics in the current scope and generate the additional Elastic kubernetes integration metrics
+	dataset := "kubernetes.node"
```
ChrsMark, line 35: Can we make those constants defined on top level?

remappers/kubernetesmetrics/kubelet.go
```diff
@@ -0,0 +1,150 @@
...
+func addKubeletMetrics(
+	src, out pmetric.MetricSlice,
+	_ pcommon.Resource,
+) error {
+	var timestamp pcommon.Timestamp
+	var total_transmited, total_received, node_memory_usage, filesystem_capacity, filesystem_usage int64
+	var cpu_limit_utilization, memory_limit_utilization, node_cpu_usage, pod_cpu_usage_node, pod_memory_usage_node float64
+
+	// iterate all metrics in the current scope and generate the additional Elastic kubernetes integration metrics
+	dataset := "kubernetes.pod"
```
ChrsMark, line 37: Same here. Making those constants helps with readability.

remappers/kubernetesmetrics/clustermetrics.go
```diff
@@ -0,0 +1,52 @@
...
+	for i := 0; i < src.Len(); i++ {
+		metric := src.At(i)
+		if metric.Name() == "k8s.node.allocatable_cpu" {
```
ChrsMark, line 38: What do you think about making these constants on top level? This would help us easily identify metric names, remappings etc on top level.

remappers/kubernetesmetrics/clustermetrics.go
```diff
@@ -0,0 +1,52 @@
...
+				timestamp = dp.Timestamp()
+			}
+			node_allocatable_memory = dp.IntValue()
+		}
+	}
+
+	remappers.AddMetrics(out, dataset, remappers.EmptyMutator,
+		remappers.Metric{
+			DataType:  pmetric.MetricTypeGauge,
+			Name:      "kubernetes.node.cpu.allocatable.cores",
```
ChrsMark, line 56: Could we define these as constants as well and maybe define the "remapping" definition in a golang map? If the map makes the result too ugly then we can just go with the constants.

gizas: Ok I moved them all in main file https://github.com/elastic/opentelemetry-lib/pull/20/files#diff-363de0029a58d415b072beaa08a95048f1ce0310f29949d1a5e1ecd4176d36afR29

Following the structure of hostmetrics.

I think now is ok

remappers/kubernetesmetrics/clustermetrics.go
gizas: See answer below

remappers/kubernetesmetrics/kubelet.go
gizas: See answer below

remappers/kubernetesmetrics/clustermetrics.go
gizas: See answer below

remappers/internal/metric.go
gizas: Ok because I see that it raises more confusion for now I copied what is in hostmetrics. 
See https://github.com/elastic/opentelemetry-lib/pull/20#discussion_r1636665657

I think that solves the diffs @lahsivjar

remappers/kubernetesmetrics/k8smetrics_test.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		if err := dp.Attributes().FromRaw(testm.DP.Attrs); err != nil {
+			t.Fatalf("failed to copy attributes from test data: %v", err)
+		}
+	}
+}
+
+func ptr[T any](v T) *T {
+	return &v
+}
```
lahsivjar, line 220: This is common to  testing all remappsers (since the metric package is common). I would suggest moving this bit into the `internal/metricstest` package or something similar and using it in both k8s and hostmetrics testing.

lahsivjar (approved):
LGTM! minor comment on making a separate test utility package.

remappers/kubernetesmetrics/clustermetrics.go
ChrsMark: It seems that only the datasets where moved not the metric names 🤔

remappers/kubernetesmetrics/k8smetrics_test.go
gizas: Thanks @lahsivjar moved this common part to internal folder.
Hostmetrics tests updated too

lahsivjar: Sorry for so much back and forth. We should put the testing in a separate package for better isolation since, _I think_, the binary would have some definitions of the structs and other things if the package is referenced in a code. Better to just create a test utility package. Feel free to merge this PR and do it as a followup.

gizas:
@lahsivjar np. Do you want me to open a new story?
Also I dont have write access to this repo. Can you please merge this PR for me if no other objections?

lahsivjar:
@gizas Yes, let's open a tracking issue for the test package.

@ishleenk17 Could you take a 👀 and merge if all good?

ishleenk17:
@gizas : Changes look inline with the hostmetrics code. 
Have we tested these changes, with the [processor](https://github.com/elastic/opentelemetry-collector-components/tree/main/processor/elasticinframetricsprocessor)?

gizas:
@ishleenk17 **not** with the elasticinframetricsprocessor (as we will need to make changes [here](https://github.com/elastic/opentelemetry-collector-components/blob/main/processor/elasticinframetricsprocessor/processor.go#L27)  is not it?) 

But we repeated the tests as mentioned [here](https://github.com/elastic/opentelemetry-lib/pull/20#issuecomment-2158755887) 

Do we need to open a PR there as well?

@lahsivjar FYI: https://github.com/elastic/opentelemetry-lib/issues/22

ishleenk17 (approved):
Please add the invocation change required in the Elasticinframetricscode as well. 
From translation perspective the changes look good. Thanks!

tetianakravchenko:
> Please add the invocation change required in the Elasticinframetricscode as well.

@gizas FYI: https://github.com/elastic/opentelemetry-collector-components/pull/14

PR #25, opened 2024-06-15
Add filesystem scraper parsing to opentelemetry-lib

ishleenk17:
Add filesystem scraper parsing to opentelemetry-lib

Relates: https://github.com/elastic/opentelemetry-dev/issues/191


remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+	"strings"
+
+	remappers "github.com/elastic/opentelemetry-lib/remappers/internal"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"golang.org/x/exp/constraints"
+)
+
+func remapFilesystemMetrics(src, out pmetric.MetricSlice,
+	_ pcommon.Resource,
```
devamanv, line 30: The `_ pcommon.Resource` parameter seems unused inside the function. Could it be removed if not needed?

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+				dp := dataPoints.At(j)
+				value := dp.IntValue()
+				timestamp = dp.Timestamp()
+				deviceValue, mpointValue, fstypeValue, ok := getAttributes(dp)
+				if !ok {
+					continue
+				}
+				device, mpoint, fstype = deviceValue.Str(), mpointValue.Str(), fstypeValue.Str()
+				// Create a unique key for each device
+				deviceKey := device + "_" + mpoint + "_" + fstype
```
devamanv, line 54:
```suggestion
				deviceKey := fmt.Sprintf("%s_%s_%s", device, mpoint, fstype)
```

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+			dataPoints := metric.Sum().DataPoints()
+			for j := 0; j < dataPoints.Len(); j++ {
+				dp := dataPoints.At(j)
+				value := dp.IntValue()
+				timestamp = dp.Timestamp()
+				deviceValue, mpointValue, fstypeValue, ok := getAttributes(dp)
+				if !ok {
+					continue
+				}
+				device, mpoint, fstype = deviceValue.Str(), mpointValue.Str(), fstypeValue.Str()
```
devamanv, line 52:
```suggestion
				device, mpoint, fstype, ok := getAttributes(dp)
				if !ok {
					continue
				}
```
I think we can directly use the returned values for `device`, `mpoint`, and `fstype` instead of explicitly converting them to string type as in L52. This was required because these were declared in L34 as strings. That shouldn't be required if they are not returned or required to be used outside the `remapFilesystemMetrics` function.

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+						}
+					}
+				}
+			}
+
+		}
+	}
+
+	for deviceKey, totalfsusage := range totalUsagePerDevice {
+		device, mpoint, fstype = parseDeviceKey(deviceKey)
```
devamanv, line 82: I am wondering what happens if deviceKey isn't a valid one? The `parseDeviceKey` function should return an error to handle it, and propagate it to the caller, which should then return it to its caller as an error. The `remapFilesystemMetrics` returns an error, but in this it will fail silently.

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+		addFileSystemMetrics(out, timestamp, dataset, "system.filesystem.total", device, mpoint, fstype, totalfsusage)
+		if usedBytes, exists := usedBytesPerDevice[deviceKey]; exists {
+			usedPercentage := float64(usedBytes) / float64(totalfsusage)
+			addFileSystemMetrics(out, timestamp, dataset, "system.filesystem.used.pct", device, mpoint, fstype, usedPercentage)
+		}
+	}
+
+	for deviceKey, totalinodeusage := range totalInodeUsagePerDevice {
+		device, mpoint, fstype = parseDeviceKey(deviceKey)
```
devamanv, line 91:
```suggestion
		device, mpoint, fstype := parseDeviceKey(deviceKey)
```
For the same reason as mentioned above for L48-L51

remappers/hostmetrics/filesystem.go
ishleenk17: This is required by remaping function in other scrapers. Might be used here as well in future.

remappers/hostmetrics/filesystem.go
ishleenk17: The device, mpoint, fstype are being used as strings in multiple places (be it within the function) while adding the metrics. Hence I would prefer keeping it this way.

devamanv: The are used as strings that's why it's better to keep them close to wherever they are being used instead of declaring them at the start of the function. Leave the choice to you.

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+
+package hostmetrics
+
+import (
+	"strings"
+
+	remappers "github.com/elastic/opentelemetry-lib/remappers/internal"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"golang.org/x/exp/constraints"
```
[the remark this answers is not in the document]
ishleenk17: @kruskall : Can you point me to the package you are referring to?

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,149 @@
...
+	for deviceKey, totalinodeusage := range totalInodeUsagePerDevice {
+		device, mpoint, fstype = parseDeviceKey(deviceKey)
+		addFileSystemMetrics(out, timestamp, dataset, "system.filesystem.files", device, mpoint, fstype, totalinodeusage)
+	}
+	return nil
+}
+
+func addFileSystemMetrics[T interface {
+	constraints.Integer | constraints.Float
+}](out pmetric.MetricSlice,
```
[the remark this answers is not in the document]
ishleenk17: Hmmm, makes sense. Let me confirm if these will always be the only types. We can use this to avoid using the constraints package.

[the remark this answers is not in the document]
shmsr: You can also create:

```
type number interface {
	~int64 | ~float64
}
```

So, it will look like `T number`; cleaner. Creating this can make it reusable across other functions, if required.

remappers/hostmetrics/filesystem.go
lahsivjar: `Sprintf` for joining only strings is a bad idea, it should give way worse performance than just plain addition.

remappers/hostmetrics/filesystem.go
lahsivjar: +1, there is no point in keeping these vars at a higher scope than the for loop.

remappers/hostmetrics/filesystem.go
lahsivjar: The whole logic of using the device key and then parsing looks like introduced complexity from us. We are first creating the device key by doing a sum and then parsing it(?) I am quite sure we should be able to merge this for loop in the first one but even if we are not we should be able to avoid the re-parsing of the key.

remappers/hostmetrics/filesystem.go
devamanv: I missed that. Agree with your point to use plain concatenation. Marking it as resolved.

remappers/hostmetrics/filesystem.go
ishleenk17: Let me try to simplify the logic

ishleenk17:
@lahsivjar : Refactored the code with below changes mainly: 
Usage of a single map 
Removal of the parsekeylogic

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,151 @@
...
+		return
+	}
+	mpoint, ok = dp.Attributes().Get("mountpoint")
+	if !ok {
+		return
+	}
+	fstype, ok = dp.Attributes().Get("type")
+
+	return
+}
```
lahsivjar, line 151:
```go
func getDeviceKey(dp pmetric.NumberDataPoint) (d deviceKey, ok bool) {
```

Can we directly return device key from this function and rename this? There is no point in keeping this generic function here.

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,151 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+)
+
+// DeviceMetrics holds metrics data for a device
+type DeviceMetrics struct {
+	TotalUsage      int64
+	UsedBytes       int64
+	TotalInodeUsage int64
+}
```
lahsivjar, line 31:
```suggestion
// DeviceMetrics holds metrics data for a device
type deviceMetrics struct {
	totalUsage      int64
	usedBytes       int64
	totalInodeUsage int64
}
```

I think we should not export this.

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,151 @@
...
+	for i := 0; i < src.Len(); i++ {
+		metric := src.At(i)
+		switch metric.Name() {
+		case "system.filesystem.usage", "system.filesystem.inodes.usage":
+			dataPoints := metric.Sum().DataPoints()
+			for j := 0; j < dataPoints.Len(); j++ {
+				dp := dataPoints.At(j)
+				value := dp.IntValue()
+				timestamp = dp.Timestamp()
+				device, mpoint, fstype, ok := getAttributes(dp)
```
lahsivjar, line 58:
```suggestion
				key, ok := getDeviceKey(dp)
```

remappers/hostmetrics/filesystem.go
```diff
@@ -0,0 +1,151 @@
...
+				if !ok {
+					continue
+				}
+				// Create a unique key for each device
+				key := deviceKey{device: device.Str(), mpoint: mpoint.Str(), fstype: fstype.Str()}
```
lahsivjar, line 63:
```suggestion
```

We can remove this given device key is retrieved before.

lahsivjar (commented):
Some slight modifications to simplify the device key creation logic a bit. Rest LGTM!

lahsivjar (approved):
Approving it now in case this is time sensitive and needs to be merged quickly. If not then would get great to get the above refactor done otherwise okay to do it in a followup PR to, I will leave it up to you.

lahsivjar (approved):
Thanks for making all the changes, LGTM!

PR #26, opened 2024-06-20
Fix issues in remappers

ishleenk17:
1. Push the processes scraper data to system.process.summary dataset
2. Add system.process.cmdline as attribute to processmetrics
3. Update the tests 
4. Rectify the name of iowait.pct in CPU scraper
5. Improve logic of calculating total number of processes

### DIFF
--- a/remappers/hostmetrics/process.go
+++ b/remappers/hostmetrics/process.go
@@ -277,5 +277,9 @@ func addProcessResources(resource pcommon.Resource) func(pmetric.NumberDataPoint
 		if name.Str() != "" {
 			dp.Attributes().PutStr("process.name", name.Str())
 		}
+		cmdline, _ := resource.Attributes().Get("process.command_line")
+		if cmdline.Str() != "" {
+			dp.Attributes().PutStr("system.process.cmdline", cmdline.Str())
+		}
 	}
 }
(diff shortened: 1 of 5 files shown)


remappers/hostmetrics/process.go
```diff
@@ -277,5 +277,9 @@ func addProcessResources(resource pcommon.Resource) func(pmetric.NumberDataPoint
 		if name.Str() != "" {
 			dp.Attributes().PutStr("process.name", name.Str())
 		}
+		cmdline, _ := resource.Attributes().Get("process.command_line")
+		if cmdline.Str() != "" {
+			dp.Attributes().PutStr("system.process.cmdline", cmdline.Str())
+		}
```
lahsivjar, line 283: I thought we were leaving the mapping of resource attributes to the callers (elatic processor, MIS/APM), is this not the case? I was under the impression that we would be getting rid of the whole of `addProcessResource`

ishleenk17: Yes, that's the plan. As of now, this doesn't exist in Exporter. Especially, handling the resource attributes whose names are different in OTEL and Elastic. Until then will keep it here. Once it is addressed in Exporter, will be removing the addProcessResource. 
The `system.process.cmdline` is being used in the Processes tab of the Host UI. Hence its important to have this.

lahsivjar (approved):

PR #31, opened 2024-06-24
Add k8s.pod.cpu.node.utilization metric in kubernetes remapper

tetianakravchenko:
The metric k8s.pod.cpu.node.utilization is available in [v0.103.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.103.0/receiver/kubeletstatsreceiver/documentation.md#k8spodcpunodeutilization)

So we open this PR to add missing metrics from the remapper


ishleenk17 (approved):
Looks good!

PR #34, opened 2024-07-02
Add required metrics and datapoint attributes for Process tab in Host UI

ishleenk17:
Relates: https://github.com/elastic/opentelemetry-dev/issues/296

Post the fix, the Process tab lights up

Screenshot 2024-07-02 at 6 30 08 PM

### DIFF
--- a/remappers/hostmetrics/process.go
+++ b/remappers/hostmetrics/process.go
@@ -290,7 +290,7 @@ func addProcessResources(resource pcommon.Resource) func(pmetric.NumberDataPoint
 			dp.Attributes().PutStr("system.process.cmdline", cmdline.Str())
 		}
 		//Adding dummy value to process.state as "Not Known", since this field is not
-		//available through Hostmetrics currently and Process tab in Curated UI's needs this field as a prerequisite to work
+		//available through Hostmetrics receiver currently and Process tab in Curated UI's need this field as a prerequisite.
 		dp.Attributes().PutStr("system.process.state", "Not Known")
 	}
 }
--- a/remappers/hostmetrics/process.go
+++ b/remappers/hostmetrics/process.go
@@ -289,8 +289,8 @@ func addProcessResources(resource pcommon.Resource) func(pmetric.NumberDataPoint
 		if cmdline.Str() != "" {
 			dp.Attributes().PutStr("system.process.cmdline", cmdline.Str())
 		}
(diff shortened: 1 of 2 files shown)


remappers/hostmetrics/process.go
```diff
@@ -281,5 +289,8 @@ func addProcessResources(resource pcommon.Resource) func(pmetric.NumberDataPoint
 		if cmdline.Str() != "" {
 			dp.Attributes().PutStr("system.process.cmdline", cmdline.Str())
 		}
+		//Adding dummy value to process.state as "Not Known", since this field is not
+		//available through Hostmetrics receiver currently and Process tab in Curated UI's need this field as a prerequisite.
+		dp.Attributes().PutStr("system.process.state", "Not Known")
```
lahsivjar, line 294: Can we use `unknown`? It is an accepted state in [otel](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processesscraper/documentation.md).

ishleenk17: I didn't want to make it similar to an existing state in OTEL, since this is a dummy value.
But, I see in the process tab that if the state is not from the ones in OTEL, it automatically puts it as unknown, irrespective of what value we give here. See the screenshot in the description.
If required I can make it Unknown here as well.

lahsivjar: Also, we might want to document this as a bug since users will get confused by seeing no processes in unknown in the summary section but seeing everything as unknown later on.

It might have been better if we could put a different name (like `undefined` or something which represents that we cannot figure it out explicitly) but I don't think UI would allow that.

lahsivjar: > If required I can make it Unknown here as well.

Fair point in keeping it different from OTel but I still think keeping `unknown` is better.

A nit: If you think keeping it separate is important then can we use `undefined` rather than a two word state.

ishleenk17: > It might have been better if we could put a different name (like undefined or something which represents that we cannot figure it out explicitly) but I don't think UI would allow that.

Correct, I would have preferred that as well. But UI doesn't allow it. Changes them to Unknown. We can list it under a "Known Issue". Will bring it up in the meeting tomorrow. 

>  then can we use undefined rather than a two word state.

Makes sense.

lahsivjar (approved):

PR #35, opened 2024-07-03
Send system.process.cpu.start_time as a field

lahsivjar:
* Converts the `system.process.cpu.start_time` to a field with string date. The string date format allows the field to be correctly detected as a `date` field rather than an `int` field by the dynamic mapping used by APM. The date mapping is required by Kibana to parse the field correctly in the Process tab.
* Adds `event.dataset` for the process summary metric which is again required by the Process tab.

### DIFF
--- a/remappers/hostmetrics/process.go
+++ b/remappers/hostmetrics/process.go
@@ -18,6 +18,8 @@
 package hostmetrics
 
 import (
+	"time"
+
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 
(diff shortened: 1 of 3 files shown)


axw (approved):

remappers/hostmetrics/process.go
```diff
@@ -289,8 +289,10 @@ func addProcessResources(resource pcommon.Resource) func(pmetric.NumberDataPoint
 		if cmdline.Str() != "" {
 			dp.Attributes().PutStr("system.process.cmdline", cmdline.Str())
 		}
-		//Adding dummy value to process.state as "undefined", since this field is not
-		//available through Hostmetrics receiver currently and Process tab in Curated UI's need this field as a prerequisite.
+		dp.Attributes().PutStr("system.process.cpu.start_time", startTimeStr)
```
ishleenk17, line 292: So, we are adding it as a string in the end.
this works fine for ES path, since we have the mapping of "date" hence the data shows up.
Do we have similar mapping in APM path as well ?

lahsivjar: We use dynamic mapping which would map a string to a date if it passes data detection ([ref](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html))

Issue #36, opened 2024-07-04
Invalid timestamp documents get generated


ishleenk17:
Currently otel-lib generates documents with invalid timestamps in some cases:

> 2024-07-01T15:50:07.366Z	error	elasticsearchexporter@v0.103.1-0.20240701121317-74244e46a6b6/elasticsearch_bulk.go:309	Drop docs: failed to index: struct { Type string "json:\"type\""; Reason string "json:\"reason\"" }{Type:"illegal_argument_exception", Reason:"the document timestamp [1970-01-01T00:00:00.000Z] is outside of ranges of currently writable indices [[2024-07-01T12:22:18.000Z,2024-07-01T16:21:07.000Z]]"}	{"kind": "exporter", "data_type": "metrics", "name": "elasticsearch", "status": 400}

Relates: https://github.com/elastic/opentelemetry-lib/issues/6 
Relates: https://github.com/elastic/opentelemetry-dev/issues/300

ishleenk17:
In otel-lib, if an OTEL metric is not present, we should not generate the corresponding translated metric.
Currently this happens:
1. If OTEL metric is not present, but timestamp has been set by some previous metric. We get a translated metric like below:

> Metric #20
Descriptor:
     -> Name: system.process.cpu.total.pct
     -> Description:
     -> Unit:
     -> DataType: Gauge
NumberDataPoints #0
Data point attributes:
     -> otel_remapped: Bool(true)
     -> data_stream.dataset: Str(system.process)
     -> process.parent.pid: Int(1)
     -> user.name: Str(ishleenkaur)
Timestamp: 2024-07-03 18:31:39.372916 +0000 UTC----> Correct timestamp
Value: 0.000000 ----> value 0


2. If OTEL metric is not present and timestamp is not set through any metric, we get a translated metric like below :

> Metric #20
Descriptor:
     -> Name: system.process.cpu.total.pct
     -> Description:
     -> Unit:
     -> DataType: Gauge
NumberDataPoints #0
Data point attributes:
     -> otel_remapped: Bool(true)
     -> data_stream.dataset: Str(system.process)
     -> process.parent.pid: Int(1)
     -> user.name: Str(ishleenkaur)
Timestamp: 1970-01-01 00:00:00 +0000 UTC. -----> Invalid timestamp
Value: 0.000000. ----> Value 0

PR #37, opened 2024-07-04
Fix timestamp issue

ishleenk17:
Closes: https://github.com/elastic/opentelemetry-lib/issues/36


remappers/internal/metric.go
```diff
@@ -46,6 +46,12 @@ func AddMetrics(
 	ms.EnsureCapacity(ms.Len() + len(metrics))
 
 	for _, metric := range metrics {
+
+		//The translated metric should not be generated for a null or negative value
+		if (metric.IntValue != nil && *metric.IntValue <= 0) || (metric.DoubleValue != nil && *metric.DoubleValue <= 0) {
```
lahsivjar, line 51:
```suggestion
		if (metric.IntValue != nil && *metric.IntValue <= 0) || (metric.DoubleValue != nil && *metric.DoubleValue <= 0 && math.IsInf(*metric.DoubleValue, 0) {
```
Should we also check for Infs here? Or is that expected to have?

ishleenk17: We can add the infinity check as an additional check.

remappers/internal/metric.go
```diff
@@ -46,6 +46,12 @@ func AddMetrics(
 	ms.EnsureCapacity(ms.Len() + len(metrics))
 
 	for _, metric := range metrics {
+
+		//The translated metric should not be generated for a null or negative value
+		if (metric.IntValue != nil && *metric.IntValue <= 0) || (metric.DoubleValue != nil && *metric.DoubleValue <= 0 && math.IsInf(*metric.DoubleValue, 0) {
```
lahsivjar, line 51:
```suggestion
		if (metric.IntValue != nil && *metric.IntValue <= 0) || (metric.DoubleValue != nil && *metric.DoubleValue <= 0 && math.IsInf(*metric.DoubleValue, 0)) {
```

lahsivjar (approved):

PR #41, opened 2024-07-05
Add event.dataset to all remapped metrics

lahsivjar:
Unlike `data_stream.dataset` which doesn't need to be set for the APM path, the `event.dataset` is required for multiple UI panels. The PR allows setting the dataset in all cases and adopts the mutator pattern throughout the codebase.

There are no major logic changes other than all metrics having `event.dataset` set as a datapoint attribute.

### DIFF
--- a/remappers/common/const.go
+++ b/remappers/common/const.go
@@ -18,6 +18,12 @@
 package common
 
 const (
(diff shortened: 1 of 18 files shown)


ishleenk17 (approved):
Changes look good!

PR #42, opened 2024-07-08
Handle user.name/system_cmdline if not present

ishleenk17:
In some scenarios, we have observed that the user.name is not fetched since ots corresponding OTEL metric(process.owner) is not present. 
But the absence of user.name leads to the process tab not working in Curated UI's. Hecne adding default value of "undefined" in case not present.
The same is done for system_cmdline for the same reasons


lahsivjar (approved):
LGTM! I am slightly concerned that this may cause some confusion with the users and we may be better off dropping such data but I guess with PID populated it would still be beneficial.

ishleenk17:
> but I guess with PID populated it would still be beneficial.

Yes, if we drop the document, no information about the process is there. With this, at least the user would know that a process with a particular pid is present in their system.

PR #43, opened 2024-07-10
Set event.module for hostmetrics remapper

lahsivjar:
`event.module` is required to make the Observability Overview panel work for the APM path.


ishleenk17 (approved):
Verified with the exporter Path. 
Works fine.

PR #46, opened 2024-07-15
Add trace enrichment logic for apm aggregation keys

lahsivjar:
Adds enrichment logic for enriching OTel spans. Only the fields required for [apm-aggregation](https://github.com/elastic/apm-aggregation) are currently added. The enrichment logic is based on https://github.com/elastic/apm-data/tree/main, however, I have tried to keep adhering to the latest semconv for now.

Related to: https://github.com/elastic/opentelemetry-dev/issues/304


ishleenk17:
@lahsivjar : So we plan to use this library for maintaining any type of compatibility between OTEL and Elastic.
Not just Metrics specific. 
That sounds fine. Shall we remove the metrics specific description in the readme of the repo then ?

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (s *Span) setTxnType(span ptrace.Span) {
+	txnType := "unknown"
+	switch {
+	case s.isMessaging:
+		txnType = "messaging"
+		span.Attributes().PutStr("transaction.type", "messaging")
+	case s.isRPC, s.isHTTP:
+		txnType = "request"
+		// TODO (lahsivjar): DB type is not handled in apm-data
```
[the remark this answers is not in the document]
lahsivjar: > There is no such things as db transaction type in Elastic APM. The [transaction represents the entry](https://github.com/elastic/apm/blob/main/specs/agents/tracing-transactions.md) to a service - that works out nicely for incoming HTTP calls and e.g. for reading out from a message queue. But e.g. a database call always happens within a transaction - that can't be the entry point into a service - something else always needs to trigger the db call - that's how the Elastic APM data model work.

Thanks for the information here. My first thoughts were that if we instrumented a DB as a service then the transaction type would be `db` but now I realize that is not correct and in such cases, the txn would be w.r.t. the db as a service.

> So... here is my question: do we maybe want to try to solve this problem?

I don't yet grasp the whole issue but I think it would be worth to solve it if we can. I will go over the apm-data issue and create one here if required. We can continue the discussion there.

lahsivjar:
@ishleenk17 I think the decision to use this library for all OTel<>Elastic compatibility was when we named it. We should definitely fix the description in the readme, I will create a followup for that.

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (s *Span) setServiceTarget(span ptrace.Span) {
+	var targetType, targetName string
+
+	if s.peerService != "" {
+		targetName = s.peerService
+	}
+
+	switch {
+	case s.isHTTP:
```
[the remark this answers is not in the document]
lahsivjar: Fixed via https://github.com/elastic/opentelemetry-lib/pull/46/commits/1d7e19708667760a5ccea344dfd655b8998fac10

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,296 @@
...
+			s.httpStatusCode = v.Int()
+		case semconv.AttributeHTTPMethod,
+			semconv.AttributeHTTPRequestMethod,
+			semconv.AttributeHTTPURL,
+			semconv.AttributeHTTPTarget,
+			semconv.AttributeHTTPScheme,
+			semconv.AttributeHTTPFlavor,
+			semconv.AttributeNetHostName:
+			s.isHTTP = true
+		// TODO (lahsivjar): Is url ONLY present for HTTP? This is what is done in apm-data.
```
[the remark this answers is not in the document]
gregkalapos: Ahh, that's the next huge topic we need to think about :) That is versioning across different SemConv versions.

`db.name` got renamed to `db.namespace` - can be [seen here](https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/model/registry/deprecated/db.yaml#L69-L74). I see Elasticsearch in v1.26.0 which you linked doesn't have `db.namespace`, but that seems to be an error - [`main` already has it](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/elasticsearch.md) - that's how it should be and `db.namespace` be there in the next version.

This happens as part of the DB stabilization - currently all db attributes are experimental, the next version aims to make most of these stable, but as part of this, there are some changes.

As it seems in SemConv v1.26.0 the renaming of `db.name` to `db.namespace` already happened (except for ES, where it just got removed) and got released in the SemConv spec, but that's just the SemConv spec, I doubt any real implementation implemented that. I'd expect most still send `db.name` - in case of Elasticsearch I'm fairly confident we haven't updated this yet (I in discussion wit the ES clients team to go through the SemConv changes and implement the new version - I'm not aware they implemented any of this yet).

> OTOH, vendor-specific attributes like, [db.elasticsearch.cluster.name](https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/attributes-registry/db.md), feel a bit brittle to handle.

Agree, that can explode quickly and the goal of SemConv is to have common attributes across vendors (and only have vendor specific ones where it's really needed), so we should take advantage of this and use the vendor agnostic attribute.

> Currently, the logic to determine service target uses db.name but if such attributes are not expected in the semconv maybe we should expand the definition.

Ultimately I think we should have a better story for version handling. In this case, depending on the version of the schema url that's sent, it should be either `db.name` (old version) or `db.namespace` from version `1.26` and higher.

To unblock the PR: since I think no instrumentation implemented `db.namespace` yet, I think having `db.name` for now is fine. And then I'd handle versioning in an extra issue/PR later. 

Alternatively you could check for both - that'd work and one should be always present.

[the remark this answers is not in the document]
lahsivjar: > db.name got renamed to db.namespace - can be [seen here](https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/model/registry/deprecated/db.yaml#L69-L74). I see Elasticsearch in v1.26.0 which you linked doesn't have db.namespace, but that seems to be an error - [main already has it](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/elasticsearch.md) - that's how it should be and db.namespace be there in the next version.

Hmm, I see. I have been working with v1.25.0 of the semconv (the latest released one with v0.104.0) so I didn't even see that db.name has been deprecated. Thanks for pointing me to it.

> To unblock the PR: since I think no instrumentation implemented db.namespace yet, I think having db.name for now is fine. And then I'd handle versioning in an extra issue/PR later.

> Alternatively you could check for both - that'd work and one should be always present.

It would probably be best to handle this in a followup PR after we have the v1.26.0 of the semconv released.

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -375,6 +375,60 @@ func TestElasticSpanEnrich(t *testing.T) {
...
+				)
+				return span
+			}(),
+			enrichedAttrs: map[string]any{
+				AttributeEventOutcome:      "success",
+				AttributeServiceTargetType: "cassandra",
+				// TODO (lahsivjar): Current apm-data would not fallback to using
+				// RPC/HTTP when a database span is detected but all fields are not
+				// populated. Should it?
+				AttributeServiceTargetName: "testsvc",
```
lahsivjar, line 429: @gregkalapos @axw What are your thoughts on this? As in this test case, we won't fallback to setting the svc target attributes in cases where span represents a db type over HTTP BUT the db attributes are not enough to infer all required information.

ishleenk17 (approved):
Code owner approval.

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,294 @@
...
+	spanStatusCode ptrace.StatusCode
+
+	isMessaging        bool
+	isRPC              bool
+	isHTTP             bool
+	isDB               bool
+	messagingDestnTemp bool
+}
+
+func (s *Span) Enrich(span ptrace.Span) {
```
axw, line 71: Given that this is the only accessible part of the API, I think it may be worth changing the API a bit to be like:

```go
func EnrichSpan(span ptrace.Span) {
	var c spanEnrichmentContext
	c.enrich(span)
}

type spanEnrichmentContext {
	urlFull *url.URL

	peerService     string
	...
}

func (c *spanEnrichmentContext) enrich(span ptrace.Span) {
	...
}
```

enrichments/trace/internal/elastic/span.go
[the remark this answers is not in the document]
axw: If we're going to do this, does it belong here? I think maybe it would belong to the same processor, but adding artificial spans to the trace seems fairly different to what we're doing in this code.

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,294 @@
...
+		s.rpcSystem = "grpc"
+	}
+}
+
+func (s *Span) setTxnType(span ptrace.Span) {
+	txnType := "unknown"
+	switch {
+	case s.isMessaging:
+		txnType = "messaging"
+		span.Attributes().PutStr("transaction.type", "messaging")
```
axw, line 171:
```suggestion
```

handled below right?

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,294 @@
...
+func (s *Span) setEventOutcome(span ptrace.Span) {
+	// default to success outcome
+	outcome := "success"
+	switch {
+	case s.spanStatusCode == ptrace.StatusCodeError:
+		outcome = "failure"
+	case s.httpStatusCode >= http.StatusInternalServerError:
+		// TODO (lahsivjar): Handle GRPC status code? - not handled in apm-data
+		// TODO (lahsivjar): Move to HTTPResponseStatusCode? Backward compatibility?
+		outcome = "failure"
```
axw, line 214: Looks like there's a slight discrepancy with apm-data here: in apm-data we only the the outcome based on HTTP status code if the span status is unset.

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeEventOutcome:      "success",
+				AttributeTransactionResult: "OK",
+				AttributeTransactionType:   "request",
+			},
+		},
+		{
+			name: "grpc_status_internal_error",
+			input: func() ptrace.Span {
+				span := ptrace.NewSpan()
+				// attributes should be preferred over span status
```
axw, line 114: This comment is a bit confusing to me: I think what you're saying is that the status code takes precedence over span status for setting the `transaction.result`? But OTOH it's the opposite for `event.outcome`. Maybe clarify that in the comment?

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeEventOutcome:      "success",
+				AttributeServiceTargetType: "http",
+				AttributeServiceTargetName: "testsvc",
+			},
+		},
+		{
+			name: "http_span_full_url",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
```
axw, line 238:
```suggestion
				// peer.service should be ignored if more specific deductions
```

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeServiceTargetType: "external",
+				AttributeServiceTargetName: "service.Test",
+			},
+		},
+		{
+			name: "messaging_basic",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
+				// can be made about the service target.
```
axw, line 334:
```suggestion
```

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeServiceTargetType: "messaging",
+				AttributeServiceTargetName: "t1",
+			},
+		},
+		{
+			name: "messaging_temp_destination",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
+				// can be made about the service target.
```
axw, line 366:
```suggestion
```

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeServiceTargetType: "grpc",
+				AttributeServiceTargetName: "testsvc",
+			},
+		},
+		{
+			name: "rpc_span_system",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
+				// can be made about the service target.
```
axw, line 302:
```suggestion
```

Does this comment belong here? It seems that `peer.service` is honoured, per the enriched attributes.

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeServiceTargetType: "http",
+				AttributeServiceTargetName: "www.foo.bar:443",
+			},
+		},
+		{
+			name: "rpc_span_grpc",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
+				// can be made about the service target.
```
axw, line 283:
```suggestion
```

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeEventOutcome:      "success",
+				AttributeServiceTargetType: "http",
+				AttributeServiceTargetName: "www.foo.bar:443",
+			},
+		},
+		{
+			name: "http_span_no_full_url",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
```
axw, line 261:
```suggestion
				// peer.service should be ignored if more specific deductions
```

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeEventOutcome:      "success",
+				AttributeServiceTargetName: "testsvc",
+			},
+		},
+		{
+			name: "http_span_basic",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
+				// can be made about the service target.
```
axw, line 220:
```suggestion
```

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -0,0 +1,447 @@
...
+				AttributeServiceTargetType: "messaging",
+				AttributeServiceTargetName: "testsvc",
+			},
+		},
+		{
+			name: "db_over_http",
+			input: func() ptrace.Span {
+				span := getElasticSpan()
+				// peer-service should be ignored if more specific deductions
+				// can be made about the service target.
```
axw, line 383:
```suggestion
```

enrichments/trace/internal/elastic/span_test.go
axw: I would be inclined to go with bug-for-bug compatibility here for now, and create an issue and follow up. I think it probably makes sense to have an order of precedence like DB > RPC > HTTP > general, both for setting the target type and target name. OTOH it's kinda awkward that one attribute would be set from DB, and another from RPC/HTTP, so it's not 100% obvious to me.

axw (commented):
Looks great overall, just a bunch of nit picks.

enrichments/trace/internal/elastic/span_test.go
lahsivjar: Copy paste fiasco 😮‍💨

enrichments/trace/internal/elastic/span_test.go
lahsivjar: Fixed! Thanks for finding that!

enrichments/trace/internal/elastic/span.go
lahsivjar: Fixed!

enrichments/trace/internal/elastic/span.go
```diff
@@ -0,0 +1,294 @@
...
+	if targetType != "" {
+		span.Attributes().PutStr(AttributeServiceTargetType, targetType)
+	}
+	if targetName != "" {
+		span.Attributes().PutStr(AttributeServiceTargetName, targetName)
+	}
+}
+
+func isElasticTransaction(span ptrace.Span) bool {
+	if span.ParentSpanID().IsEmpty() ||
```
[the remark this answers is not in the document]
lahsivjar: It would be different then what we do in apm-data right? Is it okay to create a follow-up for this?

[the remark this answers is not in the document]
gregkalapos: > We should ideally decide based on whether the span's parent is remote.

Hmm - very likely I miss something here, because I'm surprised by this. I thought this method would make the decision if this is a local transaction - so if this is the very first span (root span) in the given service. If there is no upstream service, then `isRemote` will still be `false`, but we'd still need to map that root span into a transaction.

> It would be different then what we do in apm-data right? I

Yeah, that's also [how I read it.](https://github.com/elastic/apm-data/blob/main/input/otlp/traces.go#L198)

enrichments/trace/internal/elastic/span_test.go
gregkalapos: ➕ on the suggestion. I think the core issue is that some of the fields we use here are not required (e.g. `db.name`), so this is not only a theoretical problem - this can happen while the data is 100% compatible with SemConv.

If there is an issue, I'm happy to discuss how we could improve it.

gregkalapos (approved):
Looks good to me for the points I raised. Nice PR! 👍

enrichments/trace/internal/elastic/span.go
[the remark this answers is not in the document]
lahsivjar: I have opened a followup PR for this: https://github.com/elastic/opentelemetry-lib/pull/48/commits/fe1da3a8821d5138cf65fdc6deec0ea24f3ff8d3

We can continue the discussion there.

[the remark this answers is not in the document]
axw: > If there is no upstream service, then isRemote will still be false, but we'd still need to map that root span into a transaction.

Sorry, clarifying: I meant instead of checking the span kind, rather than whether it's the root span.

axw (approved):
LGTM, thanks!

PR #61, opened 2024-07-26
Add txn root, txn name and span name attributes for trace enrichment

lahsivjar:
Related to: https://github.com/elastic/opentelemetry-dev/issues/304

### DIFF
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -161,6 +161,9 @@ func (s *spanEnrichmentContext) enrichTransaction(
 	span ptrace.Span,
 	cfg config.ElasticTransactionConfig,
 ) {
+	if cfg.Root.Enabled {
+		span.Attributes().PutBool(AttributeTransactionRoot, true)
+	}
 	if cfg.Type.Enabled {
 		s.setTxnType(span)
 	}
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -164,6 +164,9 @@ func (s *spanEnrichmentContext) enrichTransaction(
 	if cfg.Root.Enabled {
 		span.Attributes().PutBool(AttributeTransactionRoot, true)
 	}
+	if cfg.Name.Enabled {
+		span.Attributes().PutStr(AttributeTransactionName, span.Name())
+	}
 	if cfg.Type.Enabled {
 		s.setTxnType(span)
 	}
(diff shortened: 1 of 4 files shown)


enrichments/trace/internal/elastic/span.go
```diff
@@ -176,6 +182,9 @@ func (s *spanEnrichmentContext) enrichSpan(
 	span ptrace.Span,
 	cfg config.ElasticSpanConfig,
 ) {
+	if cfg.Name.Enabled {
+		span.Attributes().PutStr(AttributeSpanName, span.Name())
```
axw, line 186: Won't we already set `span.name` in the doc? Why is this needed?

lahsivjar: Span name is not an attribute and we need some way to add this to the metrics produced by the connector. The connector would only process attributes and add it to the resulting metric. We could add some support with OTTL in the connector to create attributes based on expressions but I thought this would be better since it is a kind of enrichment.

lahsivjar: For a bit more clarity, the `spanmetricsv2` connector would look like this with the current PR:

```yaml
  spanmetricsv2:
    spans:
      service_transaction:
        description: APM service transaction aggregated metrics
        attributes:
        - key: transaction.root
      transaction:
        description: APM transaction aggregated metrics
        attributes:
        - key: transaction.root
        - key: transaction.name
        - key: transaction.result
        - key: event.outcome
      span_destination:
        description: APM span destination metrics
        attributes:
          - key: span.name
          - key: event.outcome
          - key: service.target.type
          - key: service.target.name
```

axw: IMO having OTTL support would be the ideal end state. I guess this is fine for now - maybe we can come back to it.

A couple more thoughts about OTTL:

Since we only care about having `transaction.root` on transaction metrics, and nothing else, we could use the `count` connector with an OTTL-based condition on parent_span_id being empty, and (in theory) add an attribute with config in the connector, rather than in a prior enrichment processor. We would need to extend the connector with support for setting the attribute value from an OTTL expression.

axw (approved):

enrichments/trace/internal/elastic/span.go
lahsivjar: > we could use the count connector with an OTTL-based condition on parent_span_id being empty

Didn't get the count connector part since it doesn't support histograms, did you mean the span metrics connector? Also, in addition to checking for `parent_span_id` being empty, we would also need to check for remote span right?

axw: I meant theoretically, if the count connector did support histograms, we _could_ use it to produce these metrics with configuration. Something like the following (conceptually):

```yaml
  count:
    spans:
      transaction:
        description: APM transaction aggregated metrics
        conditions:
        - attributes["transaction.name"] != ""
        attributes:
        - key: transaction.root
          ottl_expression: parent_span_id == ""
        - key: transaction.name
        - key: transaction.result
        - key: event.outcome
```

> Also, in addition to checking for parent_span_id being empty, we would also need to check for remote span right?

Yes you're right, for the condition we need to check if it's a transaction, and for the value of `transaction.root` we would check the value of `parent_span_id`.

PR #63, opened 2024-07-30
[enrichments] Set svc target if either name or type is present

lahsivjar:


carsonip (approved):

enrichments/trace/internal/elastic/span.go
```diff
@@ -294,10 +294,8 @@ func (s *spanEnrichmentContext) setServiceTarget(span ptrace.Span) {
 		}
 	}
 
-	if targetType != "" {
+	if targetType != "" || targetName != "" {
```
lahsivjar, line 297: In accordance with https://github.com/elastic/apm-data/blob/a6882547d87801b212080e4d561ebdd23ad51654/input/otlp/traces.go#L1037-L1039

Issue #86, opened 2024-08-21
[enrichments] Add exception based enrichment for span events


lahsivjar:
Span events should be enriched with the following attributes:

- `error.id` (unique number for the given error instance)
- `error.exception.message`
- `error.exception.stacktrace`
- `error.exception.cause`
- `error.exception.handled`
- `error.grouping_key`

Also, add the following as enrichments for error events:
- `transaction.type`
- `transaction.sampled`

AlexanderWert:
@lahsivjar FYI: added `error.id` to the above list!

lahsivjar:
This is done! Closing!

PR #99, opened 2024-09-23
[enrichment/trace] Handle exception for span events

lahsivjar:
Related to: https://github.com/elastic/opentelemetry-lib/issues/86

### DIFF
--- a/enrichments/trace/internal/elastic/attributes.go
+++ b/enrichments/trace/internal/elastic/attributes.go
@@ -51,11 +51,7 @@ const (
 	// span event attributes
 	AttributeParentID              = "parent.id"
 	AttributeErrorID               = "error.id"
-	AttributeErrorExceptionType    = "error.exception.type"
-	AttributeErrorExceptionMessage = "error.exception.message"
 	AttributeErrorExceptionHandled = "error.exception.handled"
-	AttributeErrorStacktrace       = "error.stack_trace"
 	AttributeErrorGroupingKey      = "error.grouping_key"
 	AttributeEventKind             = "event.kind"
-	AttributeMessage               = "message"
 )
--- a/enrichments/trace/internal/elastic/attributes.go
+++ b/enrichments/trace/internal/elastic/attributes.go
@@ -53,5 +53,4 @@ const (
 	AttributeErrorID               = "error.id"
 	AttributeErrorExceptionHandled = "error.exception.handled"
 	AttributeErrorGroupingKey      = "error.grouping_key"
-	AttributeEventKind             = "event.kind"
 )
--- a/enrichments/trace/internal/elastic/attributes.go
+++ b/enrichments/trace/internal/elastic/attributes.go
@@ -47,4 +47,15 @@ const (
 	AttributeSpanDestinationServiceResource = "span.destination.service.resource"
 	AttributeSpanDurationUs                 = "span.duration.us"
 	AttributeSpanRepresentativeCount        = "span.representative_count"
+
+	// span event attributes
+	AttributeParentID              = "parent.id"
+	AttributeErrorID               = "error.id"
+	AttributeErrorExceptionType    = "error.exception.type"
+	AttributeErrorExceptionMessage = "error.exception.message"
+	AttributeErrorExceptionHandled = "error.exception.handled"
+	AttributeErrorStacktrace       = "error.stack_trace"
+	AttributeErrorGroupingKey      = "error.grouping_key"
+	AttributeEventKind             = "event.kind"
+	AttributeMessage               = "message"
 )
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -434,9 +434,6 @@ func (s *spanEventEnrichmentContext) enrich(
 	se ptrace.SpanEvent,
 	cfg config.SpanEventConfig,
 ) {
-	// TODO(lahsivjar): Is error/log context handling be done in es-exporter?
-	// Ref: https://github.com/elastic/apm-data/blob/59d33b8113d629f43dc82f9b3931ceda63a19a7a/input/otlp/traces.go#L1243-L1260
-
 	// Extract top level span event information.
 	s.exception = se.Name() == "exception"
 	if s.exception {
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -422,9 +422,8 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
 }
 
 type spanEventEnrichmentContext struct {
-	exceptionType       string
-	exceptionMessage    string
-	exceptionStacktrace string
+	exceptionType    string
+	exceptionMessage string
 
 	exception        bool
 	exceptionEscaped bool
@@ -445,8 +444,6 @@ func (s *spanEventEnrichmentContext) enrich(
 				s.exceptionType = v.Str()
 			case semconv.AttributeExceptionMessage:
 				s.exceptionMessage = v.Str()
-			case semconv.AttributeExceptionStacktrace:
-				s.exceptionStacktrace = v.Str()
 			}
 			return true
 		})
@@ -461,7 +458,6 @@ func (s *spanEventEnrichmentContext) enrich(
 	}
 	if s.exceptionType == "" && s.exceptionMessage == "" {
 		// Span event does not represent an exception
-		se.Attributes().PutStr(AttributeEventKind, "event")
 		return
 	}
 
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -422,12 +422,12 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
 }
 
 type spanEventEnrichmentContext struct {
-	exception        bool
-	exceptionEscaped bool
-
 	exceptionType       string
 	exceptionMessage    string
 	exceptionStacktrace string
+
+	exception        bool
+	exceptionEscaped bool
 }
 
 func (s *spanEventEnrichmentContext) enrich(
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -73,6 +73,8 @@ type spanEnrichmentContext struct {
 
 	spanStatusCode ptrace.StatusCode
 
+	// TODO (lahsivjar): Refactor span enrichment to better utilize isTransaction
+	isTransaction            bool
 	isMessaging              bool
 	isRPC                    bool
 	isHTTP                   bool
@@ -153,20 +155,22 @@ func (s *spanEnrichmentContext) Enrich(span ptrace.Span, cfg config.Config) {
 		return true
 	})
 
-	// Ensure all dependent attributes are handled.
 	s.normalizeAttributes()
-
-	isElasticTxn := isElasticTransaction(span)
-	if isElasticTxn {
-		s.enrichTransaction(span, cfg.Transaction)
-	} else {
-		s.enrichSpan(span, cfg.Span)
-	}
+	s.isTransaction = isElasticTransaction(span)
+	s.enrich(span, cfg)
 
 	spanEvents := span.Events()
 	for i := 0; i < spanEvents.Len(); i++ {
 		var c spanEventEnrichmentContext
-		c.enrich(spanEvents.At(i), cfg.SpanEvent)
+		c.enrich(s, spanEvents.At(i), cfg.SpanEvent)
+	}
+}
+
+func (s *spanEnrichmentContext) enrich(span ptrace.Span, cfg config.Config) {
+	if s.isTransaction {
+		s.enrichTransaction(span, cfg.Transaction)
+	} else {
+		s.enrichSpan(span, cfg.Span)
 	}
 }
 
@@ -178,7 +182,7 @@ func (s *spanEnrichmentContext) enrichTransaction(
 		span.Attributes().PutInt(AttributeTimestampUs, getTimestampUs(span.StartTimestamp()))
 	}
 	if cfg.Sampled.Enabled {
-		span.Attributes().PutBool(AttributeTransactionSampled, true)
+		span.Attributes().PutBool(AttributeTransactionSampled, s.getSampled())
 	}
 	if cfg.ID.Enabled {
 		span.Attributes().PutStr(AttributeTransactionID, span.SpanID().String())
@@ -200,7 +204,7 @@ func (s *spanEnrichmentContext) enrichTransaction(
 		span.Attributes().PutInt(AttributeTransactionDurationUs, getDurationUs(span))
 	}
 	if cfg.Type.Enabled {
-		s.setTxnType(span)
+		span.Attributes().PutStr(AttributeTransactionType, s.getTxnType())
 	}
 	if cfg.Result.Enabled {
 		s.setTxnResult(span)
@@ -252,7 +256,12 @@ func (s *spanEnrichmentContext) normalizeAttributes() {
 	}
 }
 
-func (s *spanEnrichmentContext) setTxnType(span ptrace.Span) {
+func (s *spanEnrichmentContext) getSampled() bool {
+	// Assumes that the method is called only for transaction
+	return true
+}
+
+func (s *spanEnrichmentContext) getTxnType() string {
 	txnType := "unknown"
 	switch {
 	case s.isMessaging:
@@ -260,7 +269,7 @@ func (s *spanEnrichmentContext) setTxnType(span ptrace.Span) {
 	case s.isRPC, s.isHTTP:
 		txnType = "request"
 	}
-	span.Attributes().PutStr(AttributeTransactionType, txnType)
+	return txnType
 }
 
 func (s *spanEnrichmentContext) setTxnResult(span ptrace.Span) {
@@ -430,6 +439,7 @@ type spanEventEnrichmentContext struct {
 }
 
 func (s *spanEventEnrichmentContext) enrich(
+	parentCtx *spanEnrichmentContext,
 	se ptrace.SpanEvent,
 	cfg config.SpanEventConfig,
 ) {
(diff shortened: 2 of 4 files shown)


enrichments/trace/internal/elastic/span.go
```diff
@@ -434,6 +462,43 @@ func (s *spanEventEnrichmentContext) enrich(
...
+		se.Attributes().PutStr(AttributeErrorExceptionType, s.exceptionType)
+	}
+	if cfg.ErrorExceptionMessage.Enabled {
+		se.Attributes().PutStr(AttributeErrorExceptionMessage, s.exceptionMessage)
+	}
+	if cfg.ErrorExceptionHandled.Enabled {
+		se.Attributes().PutBool(AttributeErrorExceptionHandled, !s.exceptionEscaped)
+	}
+	if cfg.ErrorStacktrace.Enabled {
+		// TODO (lahsivjar): Where to parse stacktraces?
```
lahsivjar, line 488: [**For reviewers**] In apm-data, we parse the stacktrace (only for Java) should we be doing this here to? If yes, then how should we add the parsed stacktrace?  -- I can only think of array of string where each element is a string encoded JSON representing the format expected by Elastic UI.

enrichments/trace/internal/elastic/span.go
```diff
@@ -434,6 +462,43 @@ func (s *spanEventEnrichmentContext) enrich(
...
+	if cfg.ErrorGroupingKey.Enabled {
+		// See https://github.com/elastic/apm-data/issues/299
+		hash := md5.New()
+		// ignoring errors in hashing
+		if s.exceptionType != "" {
+			io.WriteString(hash, s.exceptionType)
+		} else if s.exceptionMessage != "" {
+			io.WriteString(hash, s.exceptionMessage)
+		}
+		se.Attributes().PutStr(AttributeErrorGroupingKey, hex.EncodeToString(hash.Sum(nil)))
```
lahsivjar, line 500: [**For reviewers**] This will be updated to include the parsed stacktrace based on above resolution.

enrichments/trace/internal/elastic/span.go
```diff
@@ -417,15 +422,38 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
...
+	exceptionMessage    string
+	exceptionStacktrace string
 }
 
 func (s *spanEventEnrichmentContext) enrich(
 	se ptrace.SpanEvent,
 	cfg config.SpanEventConfig,
 ) {
+	// TODO(lahsivjar): Is error/log context handling be done in es-exporter?
+	// Ref: https://github.com/elastic/apm-data/blob/59d33b8113d629f43dc82f9b3931ceda63a19a7a/input/otlp/traces.go#L1243-L1260
```
lahsivjar, line 438: [**For reviewers**] APM-data sets context for span-events where it sets fields like `span.id`, `transaction.id`, `parent.id` based on the span ([ref](https://github.com/elastic/apm-data/blob/main/input/otlp/traces.go#L1243-L1260)), however, I am not sure if enrichments is the correct place to do this. Since the es-exporter could decide how to encode span-events into ES docs (separate or in the same doc as parent span) maybe it would be better to have this handled in the expoter. I am confused here, WDYT?

enrichments/trace/internal/elastic/span.go
```diff
@@ -417,15 +422,38 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
...
+		se.Attributes().Range(func(k string, v pcommon.Value) bool {
+			switch k {
+			case semconv.AttributeExceptionEscaped:
+				s.exceptionEscaped = v.Bool()
+			case semconv.AttributeExceptionType:
+				s.exceptionType = v.Str()
+			case semconv.AttributeExceptionMessage:
+				s.exceptionMessage = v.Str()
+			case semconv.AttributeExceptionStacktrace:
+				s.exceptionStacktrace = v.Str()
```
lahsivjar, line 452: @gregkalapos These attributes are mostly 1<>1 mappings however, I didn't see them in the otel-data plugin so added it here. Do you think we should do this here or in the otel-data plugin?

enrichments/trace/internal/elastic/attributes.go
```diff
@@ -47,4 +47,15 @@ const (
 	AttributeSpanDestinationServiceResource = "span.destination.service.resource"
 	AttributeSpanDurationUs                 = "span.duration.us"
 	AttributeSpanRepresentativeCount        = "span.representative_count"
+
+	// span event attributes
+	AttributeParentID              = "parent.id"
```
felixbarny, line 52: I wonder if we should solve that by adding a `parent.id` -> `span_id` alias in the logs data stream.

enrichments/trace/internal/elastic/attributes.go
```diff
@@ -47,4 +47,15 @@ const (
...
+	AttributeErrorID               = "error.id"
+	AttributeErrorExceptionType    = "error.exception.type"
+	AttributeErrorExceptionMessage = "error.exception.message"
+	AttributeErrorExceptionHandled = "error.exception.handled"
+	AttributeErrorStacktrace       = "error.stack_trace"
+	AttributeErrorGroupingKey      = "error.grouping_key"
+	AttributeEventKind             = "event.kind"
+	AttributeMessage               = "message"
```
felixbarny, line 60: This can be problematic as it conflicts with the [`message` -> `body_text` alias field](https://github.com/elastic/elasticsearch/blob/7b7dd91f6208251a830b28f2fb24c3ac91414fc1/x-pack/plugin/otel-data/src/main/resources/component-templates/logs-otel%40mappings.yaml#L26-L30).

enrichments/trace/internal/elastic/span.go
felixbarny: Yes, I think that all properties that are already mapped as top level fields in the [logs data stream](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/otel-data/src/main/resources/component-templates/logs-otel%40mappings.yaml) should not be handled by enrichments but in the ES exporter.

enrichments/trace/internal/elastic/span.go
```diff
@@ -434,6 +462,43 @@ func (s *spanEventEnrichmentContext) enrich(
 	if cfg.ProcessorEvent.Enabled && s.exception {
 		se.Attributes().PutStr(AttributeProcessorEvent, "error")
 	}
+	if s.exceptionType == "" && s.exceptionMessage == "" {
+		// Span event does not represent an exception
+		se.Attributes().PutStr(AttributeEventKind, "event")
+		se.Attributes().PutStr(AttributeMessage, se.Name())
```
felixbarny, line 468: I know that we're doing this in APM data, but I don't think we should set the event name as the message field.

enrichments/trace/internal/elastic/span.go
```diff
@@ -434,6 +462,43 @@ func (s *spanEventEnrichmentContext) enrich(
 	if cfg.ProcessorEvent.Enabled && s.exception {
 		se.Attributes().PutStr(AttributeProcessorEvent, "error")
 	}
+	if s.exceptionType == "" && s.exceptionMessage == "" {
+		// Span event does not represent an exception
+		se.Attributes().PutStr(AttributeEventKind, "event")
```
felixbarny, line 467: I think this should belong in the ES exporter. In the [mapping spec](https://github.com/elastic/opentelemetry-dev/blob/main/docs/design-decisions/ingest/mapping.md#span-events), we're saying that the span event name should be mapped to `attributes.event.name`. This shouldn't depend on whether or not enrichment is used.

enrichments/trace/internal/elastic/span.go
felixbarny: I think we can leave this out for now.

Alternatively, we can parse the stack trace but only use it for the error grouping hash.

lahsivjar: > Alternatively, we can parse the stack trace but only use it for the error grouping hash

Sounds good. Since we are not going to parse the stacktrace, I will work on this as a follow-up.

enrichments/trace/internal/elastic/span.go
lahsivjar: CC: @carsonip This would require some changes in the spanevent handling in es-exporter. I think just copying some attributes from the parent span should be good enough.

enrichments/trace/internal/elastic/span.go
lahsivjar: Yeah, aliases are how we have been handling such fields. Will wait for @gregkalapos to confirm

enrichments/trace/internal/elastic/span.go
lahsivjar: +1, I was quite confused with this. I will remove this.

enrichments/trace/internal/elastic/span.go
lahsivjar: CC: @carsonip

enrichments/trace/internal/elastic/span.go
carsonip: makes sense

enrichments/trace/internal/elastic/span.go
carsonip: ack

enrichments/trace/internal/elastic/span.go
gregkalapos: Yes, I'd be also for aliases as well.

One exception (no pun intended) may be the `handled` flag: from what I remember from the UI testing, the UI access that field, so it always needs to be present (We could question why, but from what I remember, that's how it is today), but [`exception.escaped`](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/#attributes) is only recommended. So we always need to populate the `handled` flag.

That's the only edge case which comes to my mind, I'd expect the other ones to work with aliases.

enrichments/trace/internal/elastic/span.go
carsonip: span event span.id, transaction.id are already done in elasticsearchexporter

enrichments/trace/internal/elastic/span.go
carsonip: attributes.event.name is already handled in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/model.go#L645-L647

enrichments/trace/internal/elastic/span.go
felixbarny: Sounds good. Could you add these aliases, Greg? One challenge will be how to add the attribute aliases just for logs, without these getting overridden with the standard set of attribute aliases. Maybe we'll need to add them to the [alias definition](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/otel-data/src/main/resources/component-templates/semconv-resource-to-ecs%40mappings.yaml) that applies to all signals.
Also, I'm not sure if we need an alias for the exception stacktrace, since we likely need to fetch that from _source anyway, because IINM, we don't have a stored field for it anyway?

enrichments/trace/internal/elastic/attributes.go
gregkalapos: I don't think we even need this field - this is really the equivalent of `body_text`, isn't it? 

So, 1) it's either an [exception on span](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/) - I don't think there is any way to populate this field (top level message) via that API (it'll have exception.message in the attribute, but not on top level), 

or 

2) it's an [exception in logs](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/). That doc calls out that:
> Exceptions SHOULD be recorded as attributes on the [LogRecord](https://github.com/open-telemetry/opentelemetry-specification/tree/v1.35.0/specification/logs/data-model.md#log-and-event-record-definition)

and LogRecord has the body as a top level field - so if this is a log enriched with exception info, then we already have the field to store the message.

I mean, ultimately  both cases end up being a LogRecord, which then has the `body` field.

enrichments/trace/internal/elastic/span.go
gregkalapos: But we don't even need the aliases, don't we? 🤔 This should all work out via `passthrough`. 

Following could be sent via attributes and would be visible top level:
- `exception.type`
- `exception.message`
( above 2 are in SemConv - I don't think we even need to do anything with those here in the enrichment lib)
- `exception.handled`: that is not part of OTel, se we need to set it here.

I opened a draft: https://github.com/elastic/elasticsearch/pull/113636 I commented out all the aliases, the test is still green which queries `exception.type` top level.

lahsivjar: @gregkalapos I think the alias should be for `error.exception.{type, message}` -> `exception.{type, message}`

lahsivjar: Also, I think we might need to put alias for stacktrace too: `error.stack_trace` -> `exception.stacktrace`

enrichments/trace/internal/elastic/attributes.go
lahsivjar: > I'm afraid the "parent" in parent.id is not the span, but the parent of span, i.e. parent.id != span_id.

I don't think this is true for errors. In case of errors, we specifically override the parent ID of the `SpanEvent` to be the `SpanID` (or `TransactionID for elastic transaction): https://github.com/elastic/apm-data/blob/bd5e34116889942c53b186deab0f76beaf182de0/input/otlp/traces.go#L1255-L1259

enrichments/trace/internal/elastic/span.go
lahsivjar: Removed in https://github.com/elastic/opentelemetry-lib/pull/99/commits/4361255189c752b09bf2bed60016060031fdf858

enrichments/trace/internal/elastic/attributes.go
lahsivjar: Removed `message` in https://github.com/elastic/opentelemetry-lib/pull/99/commits/4361255189c752b09bf2bed60016060031fdf858

enrichments/trace/internal/elastic/span.go
lahsivjar: @carsonip Not sure if I am reading something incorrectly but I don't see all the fields handled in the exporter: https://github.com/elastic/apm-data/blob/59d33b8113d629f43dc82f9b3931ceda63a19a7a/input/otlp/traces.go#L1243-L1273, for example transaction type.

carsonip: transaction.type is not defined in mapping spec: https://github.com/elastic/opentelemetry-dev/blob/main/docs/design-decisions/ingest/mapping.md , that's why es exporter did not have specific handling for them. Do we actually need transaction.type for span events? If so, should they be in attributes.* or at top level?

felixbarny: > Do we actually need transaction.type for span events?

I don't know

> If so, should they be in attributes.* or at top level?

If needed, it should be under attributes.

enrichments/trace/internal/elastic/span.go
felixbarny: I actually meant to comment on the line below, sorry for the confusion. The `event.kind` is ECS specific (https://www.elastic.co/guide/en/ecs/current/ecs-allowed-values-event-kind.html), so this shouldn't be handled in the exporter. Ideally wouldn't add this at all if possible. Is the UI relying on this?

enrichments/trace/internal/elastic/span.go
gregkalapos: Oh, [I see](https://github.com/elastic/kibana/blob/2ec24f2f48b67a1c928fa6edb84214de57b59dd2/packages/kbn-apm-types/src/es_fields/apm.ts#L108-L111).

Ok, we can do `error.exception.{type, message}` -> `exception.{type, message}`; I'll update the PR.

Same for stack trace.

enrichments/trace/internal/elastic/span.go
lahsivjar: > so this shouldn't be handled in the exporter

Ah, makes sense.

> Ideally wouldn't add this at all if possible. Is the UI relying on this?

I am not sure, when creating this PR I did a crude search on Kibana repo and found some usage of `event.kind` but I have also [asked on slack](https://elastic.slack.com/archives/C07K6S1GDJB/p1727456600867439) if it is used. Since this is supposed to be in ECS if required, I think we can unblock the PR. I will remove this if the UI team confirm that it is not used.

lahsivjar: Confirmed on slack that `event.kind` is not required so I have removed it from the PR.

enrichments/trace/internal/elastic/attributes.go
carsonip: oh wow, I missed that. Then is it a bug for logs from span events?

enrichments/trace/internal/elastic/span.go
carsonip: > Yes, I think that all properties that are already mapped as top level fields in the [logs data stream](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/otel-data/src/main/resources/component-templates/logs-otel%40mappings.yaml) should not be handled by enrichments but in the ES exporter.

> If needed, it should be under attributes.

As they are under attributes, it is better done in enrichment

enrichments/trace/internal/elastic/attributes.go
carsonip: does this fix in apm-data look right? https://github.com/elastic/apm-data/pull/374
If so, yes, I think we can have `parent.id` alias -> `span.id`.

carsonip: we are discussing on slack the alternative to make the UI use `span.id` directly: https://elastic.slack.com/archives/C95SB62AG/p1727693736192519?thread_ts=1727462154.365409&cid=C95SB62AG

enrichments/trace/internal/elastic/span.go
lahsivjar: > As they are under attributes, it is better done in enrichment

I don't think it is a good idea to do this as enrichments since the fields we are talking about here are derived from the span. The ES exporter decides HOW to extract and encode the span event (in a separate doc or in the same doc). Whlie currently we are encoding them as a separate doc making this dependent on enrichment doesn't feel correct to me.

(https://github.com/elastic/opentelemetry-lib/pull/99#discussion_r1772810837 discussion thread is also related to the same topic)

lahsivjar: Me and @carsonip talked about this IRL and decided to do it as enrichments for now due to the differences in the handling of error and log contexts in apm-data ([ref](https://github.com/elastic/apm-data/blob/bd5e34116889942c53b186deab0f76beaf182de0/input/otlp/traces.go#L1243-L1273)). Once the specs are a bit more clearer around this we can revisit the enrichment.

lahsivjar: > https://github.com/elastic/opentelemetry-lib/pull/99/commits/222386e688fef526a691c98c0eb0569569c9e255

The above commit adds `transaction.{type, sampled}` enrichment for span event.

carsonip (approved):

felixbarny (approved):

PR #108, opened 2024-10-10
[enrichments]Allow overriding host name with k8s node name

lahsivjar:
// Host name is set same as k8s node name. In case, both host name
	// and k8s node name are set then host name is overridden as this is
	// considered an invalid configuration/smell and k8s node name is
	// given higher preference.

For more context: https://elastic.slack.com/archives/C07Q1CVD9MW/p1728487638317189


rogercoll (approved):
LGTM

PR #110, opened 2024-10-11
Support service.{address, port} with fallback to net.peer.*

lahsivjar:
`service.{address, port}` is used to calculate service destination when `url.*` attributes are not present. The logic fallback to use the now deprecated `net.peer.{name, port}` if `service.{address, port}` are not present.

Related to: https://github.com/elastic/opentelemetry-dev/issues/490

### DIFF
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -611,9 +611,6 @@ func getValueForKeyInString(str string, key string, separator rune, assignChar r
 	return ""
 }
 
-// getHostPort derives the host:port value from url.* attributes. Unlike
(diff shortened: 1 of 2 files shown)


enrichments/trace/internal/elastic/span.go
```diff
@@ -588,15 +614,23 @@ func getValueForKeyInString(str string, key string, separator rune, assignChar r
 // getHostPort derives the host:port value from url.* attributes. Unlike
 // apm-data, the current code does NOT fallback to net.* or http.*
 // attributes as most of these are now deprecated.
```
gregkalapos, line 616: The comment for this method is deprecated (pun intended). 

```suggestion
```

PR #116, opened 2024-10-28
[enrichments] Add support for inferred spans

lahsivjar:
Adds support for inferred spans, check [agent specs](https://github.com/elastic/apm/blob/main/specs/agents/tracing-spans.md#inferred-spans) for more details.

### DIFF
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -465,20 +465,21 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
 func (s *spanEnrichmentContext) setInferredSpans(span ptrace.Span) {
 	spanLinks := span.Links()
 	childIDs := pcommon.NewSlice()
-	for i := 0; i < spanLinks.Len(); i++ {
-		spanLink := spanLinks.At(i)
+	spanLinks.RemoveIf(func(spanLink ptrace.SpanLink) (remove bool) {
 		spanID := spanLink.SpanID()
-		spanLink.Attributes().RemoveIf(func(k string, v pcommon.Value) bool {
+		spanLink.Attributes().Range(func(k string, v pcommon.Value) bool {
 			switch k {
 			case "is_child", "elastic.is_child":
 				if v.Bool() && !spanID.IsEmpty() {
+					remove = true // remove the span link if it has the child attrs
 					childIDs.AppendEmpty().SetStr(hex.EncodeToString(spanID[:]))
 				}
-				return true
+				return false // stop the loop
 			}
-			return false
+			return true
 		})
-	}
+		return remove
+	})
 
 	if childIDs.Len() > 0 {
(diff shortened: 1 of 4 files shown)


enrichments/trace/internal/elastic/span.go
```diff
@@ -456,6 +462,29 @@ func (s *spanEnrichmentContext) setDestinationService(span ptrace.Span) {
...
+func (s *spanEnrichmentContext) setInferredSpans(span ptrace.Span) {
+	spanLinks := span.Links()
+	childIDs := pcommon.NewSlice()
+	for i := 0; i < spanLinks.Len(); i++ {
+		spanLink := spanLinks.At(i)
+		spanID := spanLink.SpanID()
+		spanLink.Attributes().RemoveIf(func(k string, v pcommon.Value) bool {
+			switch k {
+			case "is_child", "elastic.is_child":
+				if v.Bool() && !spanID.IsEmpty() {
```
lahsivjar, line 474: [**For reviewers**] In APM-Server we add an empty string in the slice if span ID is empty ([ref](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/traces.go#L1299)), however, I didn't think it was important so didn't handle it here. Let me know if it needs to be added.

lahsivjar: > I don't think span links are allowed to have an empty span ID in OTLP?

I am not sure about this, I kept the span ID check for now, let me know if you think we should remove this.

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -319,20 +321,67 @@ func TestElasticTransactionEnrich(t *testing.T) {
...
+			// Merge with the expected attributes and override the span links.
 			for k, v := range tc.enrichedAttrs {
-				expectedAttrs[k] = v
+				expectedSpan.Attributes().PutEmpty(k).FromRaw(v)
+			}
+			// Override span links
+			if tc.expectedSpanLinks != nil {
+				tc.expectedSpanLinks.CopyTo(expectedSpan.Links())
+			} else {
+				expectedSpan.Links().RemoveIf(func(_ ptrace.SpanLink) bool { return true })
```
lahsivjar, line 378: [**For reviewers**] This piece is a bit weird because CopyTo doesn't handle non initialized empty span links.

enrichments/trace/internal/elastic/span.go
JonasKunz: > In APM data we add an empty string in the slice if span ID is empty ([ref](https://github.com/elastic/apm-data/blob/46f8e38486b772090fbf1fdccc8eeea13c513075/input/otlp/traces.go#L1299)), however, I didn't think it was important so didn't handle it here

Just wanted to agree with you here. I'd keep the span ID check as is just as a safety measure.

JonasKunz (approved):
LGTM

PR #125, opened 2024-11-28
Enrich root spans that represent a dependency

gregkalapos:
In the OTel data model it's allowed (and not very uncommon) to have a root span which in itself represents a dependency. Such spans are root and exit span at the same time.

Examples can be seen in the tests (e.g. outgoing HTTP call, just a span without root that represents a db call or produces a message). Such spans are mapped into a transaction, because in the Elastic data model we always need a transaction as root.

This PR adds logic to enrich such transactions with the necessary fields to calculate dependencies and draw a service map.

Follow up PR in Kibana will still need to be done to query such transactions - currently only `processor.event: span` is used for dependencies. This PR adds all the necessary fields, so then those queries can be updated and filtering on `processor.event` can be removed (or explicitly filter to both spans and transaction in case removing is not needed).

### DIFF
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -196,6 +196,7 @@ func (s *spanEnrichmentContext) Enrich(span ptrace.Span, cfg config.Config) {
 func (s *spanEnrichmentContext) enrich(span ptrace.Span, cfg config.Config) {
 	if s.isTransaction {
 		s.enrichTransaction(span, cfg.Transaction)
+		s.enrichExitSpanTransaction(span, cfg)
 	} else {
 		s.enrichSpan(span, cfg.Span)
 	}
@@ -241,6 +242,39 @@ func (s *spanEnrichmentContext) enrichTransaction(
 	}
 	if cfg.InferredSpans.Enabled {
 		s.setInferredSpans(span)
+	}
+}
+
+// In OTel a root span can represent an outgoing call or a producer span
+// In such cases, the span is still mapped into a transaction, but such spans are enriched
+// with additional attributes that are specific to the outgoing call or producer span.
+func (s *spanEnrichmentContext) enrichExitSpanTransaction(
+	span ptrace.Span,
+	cfg config.Config,
+) {
+	if span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer {
+		if cfg.Span.TypeSubtype.Enabled {
+			s.setSpanTypeSubtype(span)
+		}
+		if cfg.Span.ServiceTarget.Enabled {
+			s.setServiceTarget(span)
+		}
+		if cfg.Span.DestinationService.Enabled {
+			s.setDestinationService(span)
+		}
+		if cfg.Span.Name.Enabled {
+			span.Attributes().PutStr(AttributeSpanName, span.Name())
+		}
+		if cfg.Transaction.Type.Enabled {
+			spanTypeAttr, hasType := span.Attributes().Get(AttributeSpanType)
+			if hasType {
+				transactionType := spanTypeAttr.Str()
+				if spanSubtypeAttr, hasSubType := span.Attributes().Get(AttributeSpanSubtype); hasSubType {
+					transactionType += "." + spanSubtypeAttr.Str()
+				}
+				span.Attributes().PutStr(AttributeTransactionType, transactionType)
+			}
+		}
 	}
 }
 
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -251,39 +251,6 @@ func (s *spanEnrichmentContext) enrichTransaction(
 	}
 }
 
-// In OTel, a local span can represent an outgoing call or a producer span.
-// In such cases, the span is still mapped into a transaction, but enriched
-// with additional attributes that are specific to the outgoing call or producer span.
-func (s *spanEnrichmentContext) enrichExitSpanTransaction(
-	span ptrace.Span,
-	cfg config.Config,
-) {
-	if span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer {
-		if cfg.Span.TypeSubtype.Enabled {
-			s.setSpanTypeSubtype(span)
-		}
-		if cfg.Span.ServiceTarget.Enabled {
-			s.setServiceTarget(span)
-		}
-		if cfg.Span.DestinationService.Enabled {
-			s.setDestinationService(span)
-		}
-		if cfg.Span.Name.Enabled {
-			span.Attributes().PutStr(AttributeSpanName, span.Name())
-		}
-		if cfg.Transaction.Type.Enabled {
-			spanTypeAttr, hasType := span.Attributes().Get(AttributeSpanType)
-			if hasType {
-				transactionType := spanTypeAttr.Str()
-				if spanSubtypeAttr, hasSubType := span.Attributes().Get(AttributeSpanSubtype); hasSubType {
-					transactionType += "." + spanSubtypeAttr.Str()
-				}
-				span.Attributes().PutStr(AttributeTransactionType, transactionType)
-			}
-		}
-	}
-}
-
 func (s *spanEnrichmentContext) enrichSpan(
 	span ptrace.Span,
 	cfg config.Config,
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -256,13 +256,13 @@ func (s *spanEnrichmentContext) enrichSpan(
 	cfg config.Config,
 	isExitRootSpan bool,
 ) {
-	if cfg.TimestampUs.Enabled {
+	if cfg.Span.TimestampUs.Enabled {
 		span.Attributes().PutInt(elasticattr.TimestampUs, getTimestampUs(span.StartTimestamp()))
 	}
-	if cfg.Name.Enabled {
+	if cfg.Span.Name.Enabled {
 		span.Attributes().PutStr(elasticattr.SpanName, span.Name())
 	}
-	if cfg.RepresentativeCount.Enabled {
+	if cfg.Span.RepresentativeCount.Enabled {
 		repCount := getRepresentativeCount(span.TraceState().AsRaw())
 		span.Attributes().PutDouble(elasticattr.SpanRepresentativeCount, repCount)
 	}
@@ -272,7 +272,7 @@ func (s *spanEnrichmentContext) enrichSpan(
 	if cfg.Span.EventOutcome.Enabled {
 		s.setEventOutcome(span)
 	}
-	if cfg.DurationUs.Enabled {
+	if cfg.Span.DurationUs.Enabled {
 		span.Attributes().PutInt(elasticattr.SpanDurationUs, getDurationUs(span))
 	}
 	if cfg.Span.ServiceTarget.Enabled {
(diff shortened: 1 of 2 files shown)


enrichments/trace/internal/elastic/span.go
```diff
@@ -237,6 +238,39 @@ func (s *spanEnrichmentContext) enrichTransaction(
...
+		}
+		if cfg.Span.Name.Enabled {
+			span.Attributes().PutStr(AttributeSpanName, span.Name())
+		}
+		if cfg.Transaction.Type.Enabled {
+			spanTypeAttr, hasType := span.Attributes().Get(AttributeSpanType)
+			if hasType {
+				transactionType := spanTypeAttr.Str()
+				if spanSubtypeAttr, hasSubType := span.Attributes().Get(AttributeSpanSubtype); hasSubType {
+					transactionType += "." + spanSubtypeAttr.Str()
```
gregkalapos, line 266: I'd be happy to hear opinion on this one.

The APM UI groups transactions based on transaction type. The idea here is that for such exit&root spans we'd want to use a different transaction type so they don't end up in the same group with incoming transactions.

Screenshot 2024-11-28 at 16 55 18

E.g. the default transaction type for incoming HTTP is `request` - with this, for outgoing we'd use `external.http`. It can be that a service sends `GET` both as outgoing root span and as an incoming transaction. This way those 2 transactions will have different types.

Alternatively to `.` we could also have `/` - e.g. `external/http`.

enrichments/trace/internal/elastic/span.go
```diff
@@ -189,6 +189,7 @@ func (s *spanEnrichmentContext) Enrich(span ptrace.Span, cfg config.Config) {
 func (s *spanEnrichmentContext) enrich(span ptrace.Span, cfg config.Config) {
 	if s.isTransaction {
 		s.enrichTransaction(span, cfg.Transaction)
+		s.enrichExitSpanTransaction(span, cfg)
```
felixbarny, line 192: Feel free to ignore, maybe this is personal preference. But the code may be a bit easier to follow when pulling in the if condition to this method, so that it's clear that only transactions that are of type client or producer get enriched.

enrichments/trace/internal/elastic/span.go
```diff
@@ -237,6 +238,39 @@ func (s *spanEnrichmentContext) enrichTransaction(
 	}
 }
 
+// In OTel a root span can represent an outgoing call or a producer span
```
felixbarny, line 241:
```suggestion
// In OTel, a local root span can represent an outgoing call or a producer span
```

enrichments/trace/internal/elastic/span.go
```diff
@@ -237,6 +238,39 @@ func (s *spanEnrichmentContext) enrichTransaction(
...
+// In such cases, the span is still mapped into a transaction, but such spans are enriched
```
felixbarny, line 242:
```suggestion
// In such cases, the span is still mapped into a transaction, but enriched
```

felixbarny (approved):

enrichments/trace/internal/elastic/span.go
AlexanderWert: will that impact other things like service map or so?

enrichments/trace/internal/elastic/span.go
```diff
@@ -237,6 +238,39 @@ func (s *spanEnrichmentContext) enrichTransaction(
...
+// with additional attributes that are specific to the outgoing call or producer span.
+func (s *spanEnrichmentContext) enrichExitSpanTransaction(
+	span ptrace.Span,
+	cfg config.Config,
+) {
+	if span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer {
```
AlexanderWert, line 248: I'm not sure we can rely on that condition, tbh.

Since that entire situation we try to handle here is an instrumentation bug in itself (IN THEORY: there never should be a single span that is entry and exit at the same time, but IN PRACTICE that happens in some situations), I don't think we should rely on `SpanKind` at all, tbh.

For example, what if the SpanKind == `SpanKindServer`, but the span still represents entry and exit at the same time?

How about, we use heuristics similar to how we do it for deriving `span.type` to identify whether a span is an exit span or not. For example, checking for certain attributes that would only make sense on exit spans but not on entry spans.

enrichments/trace/internal/elastic/span.go
AlexanderWert: I'm wondering whether we need a dedicated method for this at all.

Couldn't we just have some heuristic method in addition that just identifies whether a span is an exist span, use that in an additional condition here, but then just calling the existing `enrichSpan()` method.

enrichments/trace/internal/elastic/span.go
```diff
@@ -237,6 +238,39 @@ func (s *spanEnrichmentContext) enrichTransaction(
...
+		if cfg.Span.TypeSubtype.Enabled {
+			s.setSpanTypeSubtype(span)
+		}
+		if cfg.Span.ServiceTarget.Enabled {
```
AlexanderWert, line 252: @gregkalapos What about the thing we discussed about having two `ProcessorEvent` values in this case?

gregkalapos: I tested it and that triggers an error when the water-flow chart is generated. These transactions will be also returned for all queries that look for spans and there were some missing span related fields there.

@axw had a related idea: we could just remove the filter on `processor.event` from the relevant queries in Kibana. I'd rather explore that option and potentially get rid of `processor.event` in those queries.

enrichments/trace/internal/elastic/span.go
gregkalapos: This is a valid point and was thinking about this as well. My idea was to be on the safe side and only enrich real exit spans with the necessary attributes.

I understand we have this issue with nginx proxy spans that are a single span for both incoming and outgoing calls. I don't know what's the span type there, but if it's `SpanKindServer` then we would not enrich that indeed.

But that use-case is already a bug and I'm not sure how much we want to unwind such bugs.

> For example, what if the SpanKind == SpanKindServer, but the span still represents entry and exit at the same time?

I think that's bug in the instrumentation. [SemConv says ](https://opentelemetry.io/docs/specs/semconv/http/http-spans/):
> For an HTTP client span, SpanKind MUST be Client.

and 

> For an HTTP server span, SpanKind MUST be Server.

So for HTTP that goes against the spec and I'm not sure how much we should fight these bugs.

> How about, we use heuristics similar to how we do it for deriving span.type to identify whether a span is an exit span or not. For example, checking for certain attributes that would only make sense on exit spans but not on entry spans.

We could explore that, but I'm not sure how reliable it is. The check if something is an exit span is a bit different - there is no differentiation in that check between incoming and outgoing spans, because incomings are already assumed to be transactions, so all those cases are already filtered out at that point by only doing the check for spans.

I just quickly looked at HTTP spans, what I see, the server side only has `url.path` which is only required in server spans, and is not present in client spans. And then `url.full` is only on theclient-side. But if spankind is already invalid, I'm not sure how much we can rely on attributes. E.g. what if both are present?

Overall I feel we'd end up with a messy heuristic which may not work in all cases and may not even unwind these bugs in a way we'd want to.

If we relax this check here, I think the risk we run here is that we may categorize some incoming calls as exit spans and we start calculating dependencies for those. That'd be very bad. If we can come up with something to avoid that, then I'd feel more confident to relax this.

enrichments/trace/internal/elastic/span.go
gregkalapos: So far, I haven't seen any impact. Service map is mostly driven by the `target.*` and `destination.*` fields from what I know.

The dependency list has icons - that is based on this span, but that is not driven by `transaction.type` either, so that's not impacted either.

enrichments/trace/internal/elastic/span.go
gregkalapos: I still needed to extend `enrichSpan()` a bit, but now I got rid of `enrichExitSpanTransaction` - so there is only `enrichSpan` and `enrichTransaction` as we had it originally.

The 2 main differences in `enrichExitSpanTransaction` where:
- Not setting `processor.event` to `span`
- Setting `transaction.type`

We handle these now with a condition.

enrichments/trace/internal/elastic/span.go
gregkalapos: Took the change, but moved this comment.

enrichments/trace/internal/elastic/span.go
```diff
@@ -187,10 +187,17 @@ func (s *spanEnrichmentContext) Enrich(span ptrace.Span, cfg config.Config) {
 }
 
 func (s *spanEnrichmentContext) enrich(span ptrace.Span, cfg config.Config) {
+
+	// In OTel, a local root span can represent an outgoing call or a producer span.
+	// In such cases, the span is still mapped into a transaction, but enriched
+	// with additional attributes that are specific to the outgoing call or producer span.
+	isExitRootSpan := s.isTransaction && span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer
```
gregkalapos, line 194: This check is still up for discussion: https://github.com/elastic/opentelemetry-lib/pull/125#discussion_r1863663594

enrichments/trace/internal/elastic/span.go
gregkalapos: Discussed this with @AlexanderWert and we said, the correct way would be to still set both `span` and `transaction` and it'd be nice to not have to update Kibana.

In https://github.com/elastic/opentelemetry-lib/pull/125/commits/3def749edba2df55722e7deaba60e0d0521c3821 I changed this.

Now that we share the `enrichSpan()` method, there are no missing field errors when the waterfall chart is generated. With above commit, this seems working:

Screenshot 2024-12-09 at 19 28 32

However, order matters here. [This is the code in Kibana](https://github.com/elastic/kibana/blob/main/x-pack/plugins/observability_solution/apm/public/components/app/transaction_details/waterfall_with_summary/waterfall_container/waterfall/waterfall_helpers/waterfall_helpers.ts#L299-L308) that builds up the chart - and if it's `[span transaction]`, then it's still categorized as a span and not as a transaction leading to an empty waterfall chart.

So, it seems we can set both `span` and `transaction` on such root exit spans, but I'd say mid/long term we should still update Kibana to make it more robust.

gregkalapos: Ended up reverting this part - this is too risky.

Even if we can properly store both values in the field, Kibana queries still rely on the order of those values - if the value is in the reverse order, the UI breaks.

My plan currently is this: we will continue using `processor.event` as a single value field and I'll open PRs for the Kibana queries to not rely on this field when dependencies are calculated (or if that's not possible, then instead of just fetch `span`, fetch both `span` and `transaction`.

enrichments/trace/internal/elastic/span.go
```diff
@@ -246,39 +253,51 @@ func (s *spanEnrichmentContext) enrichTransaction(
 
 func (s *spanEnrichmentContext) enrichSpan(
 	span ptrace.Span,
-	cfg config.ElasticSpanConfig,
+	cfg config.Config,
```
lahsivjar, line 256: **nit**: I would prefer if we took `ElasticSpanConfig` and a boolean for `transactionTypeEnabled` rather than the whole config to prevent future smells with mixing the two configs.

enrichments/trace/internal/elastic/span.go
```diff
@@ -246,39 +253,51 @@ func (s *spanEnrichmentContext) enrichTransaction(
...
 	}
+	if cfg.Span.ProcessorEvent.Enabled && !isExitRootSpan {
+		span.Attributes().PutStr(elasticattr.ProcessorEvent, "span")
+	}
+
+	if isExitRootSpan && cfg.Transaction.Type.Enabled {
+		spanTypeAttr, hasType := span.Attributes().Get(elasticattr.SpanType)
+		if hasType {
+			transactionType := spanTypeAttr.Str()
+			if spanSubtypeAttr, hasSubType := span.Attributes().Get(elasticattr.SpanSubtype); hasSubType {
```
lahsivjar, line 295: **nit**: Get will iterate over the whole slice, how about updating the `setSpanTypeSubtype` to return `(string, string)` (type and subtype respectively) and. use that directly?

enrichments/trace/internal/elastic/span.go
gregkalapos: Makes sense - adapted it.

enrichments/trace/internal/elastic/span.go
gregkalapos: Thanks, agreed - pushed a change with this

enrichments/trace/internal/elastic/span.go
lahsivjar: https://github.com/elastic/opentelemetry-lib/pull/125#discussion_r1863711422 makes sense to me too

lahsivjar (approved):
Code LGTM! Regarding the pending discussion on https://github.com/elastic/opentelemetry-lib/pull/125#discussion_r1863663594, I have added a comment.

enrichments/trace/internal/elastic/span.go
gregkalapos: Created https://github.com/elastic/opentelemetry-lib/issues/162 to discuss this. 

The PR in this form will address the spec conform root spans. I think on this part we agree and I believe this is already fairly valuable. For special cases we can continue in #162 and I can open a follow up PR with just refining the check for `isExitRootSpan`.

PR #127, opened 2024-12-03
Set GenAi subtype for spans that have the gen_ai.system attribute

### LINKED ISSUE #111: unknown type for gen_ai spans
When tracing a python langchain application with opentelemetry-instrumentation-langchain (provided by openllmetry) we may produce spans like this:

```
{
    "name": "AzureChatOpenAI.chat",
    "context": {
        "trace_id": "0xdf0874b5202669e839de54311930359c",
        "span_id": "0x731e8d3b3a82268f",
        "trace_state": "[]"
    },
    "kind": "SpanKind.CLIENT",
    "parent_id": "0xdf313e3a840d43d3",
    "start_time": "2024-10-15T08:04:04.522445Z",
    "end_time": "2024-10-15T08:04:06.125744Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "traceloop.workflow.name": "ConversationalRetrievalChain",
        "traceloop.entity.path": "StuffDocumentsChain.LLMChain",
        "gen_ai.system": "Langchain",
        "llm.request.type": "chat",
        "gen_ai.request.model": "gpt-4o-mini",
        "gen_ai.response.model": "gpt-4o-mini",
        "gen_ai.request.temperature": 0.7,
        "gen_ai.prompt.0.role": "user",
        "gen_ai.prompt.0.content": "somecontent",
        "gen_ai.completion.0.role": "assistant"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "telemetry.sdk.language": "python",
            "telemetry.sdk.name": "opentelemetry",
            "telemetry.sdk.version": "1.27.0",
            "service.name": "foobar",
            "service.version": "1.0",
            "deployment.environment": "production",
            "service.instance.id": "gcpvmlls01",
            "telemetry.auto.version": "0.48b0"
        },
        "schema_url": ""
    }
}
```

that ends up having an `unknown` span type in kibana which does not look great.

Image

Please note that this span is not following the OTel genai convention but if we want to add a special type for genai stuff the presence of the  `gen_ai.system` attribute is in the semconv while this particular value (langchain) is not.

Here's the current enrichment code setting the span.type:
https://github.com/elastic/opentelemetry-lib/blob/5f809d469ae989000343dd939b2271fb46c39cf2/enrichments/trace/internal/elastic/span.go#L342-L372

xrmx:
Would like to avoid spans with unknown as type in kibana for genai libraries traced calls.

Schermata del 2024-12-03 14-45-57

This has only been unit tested here, no clue on how to run this :) 

Closes #111


gregkalapos (approved):
I think this makes sense.

Only reason this wasn't here is that this logic pre-dates popularity of genai.

Other side note: this code will be used within the OTel collector, but won't be used in APM Server/MIS, right?

I'd say that's fine, just flagging this, because until now, EDOT collector and APM Server/MIS behaved very similarly and here they'll diverge, which may be strange for users.

enrichments/trace/internal/elastic/span.go
```diff
@@ -361,6 +370,9 @@ func (s *spanEnrichmentContext) setSpanTypeSubtype(span ptrace.Span) {
 	case s.isHTTP:
 		spanType = "external"
 		spanSubtype = "http"
+	case s.isGenAi:
+		spanType = "external"
+		spanSubtype = "genai"
```
trentm, line 375: Did you consider `spanType="genai"` and `spanSubtype=s.genaiSystem`?

Also possible debate on `genai` vs `gen_ai`.

trentm: I'm not sure how the Kibana UI is using span.type vs span.subtype there.

gregkalapos: For db it shows the db system, which is the subtype:
Screenshot 2024-12-03 at 16 49 30

So, I think here it aways shows the subtype. 

`gen_ai.system` has a very limited set of well known values: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/gen-ai.md

But then SemConv says:
> If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

That last part with custom value puts me off a bit.

But no strong opinion from my side either. I think both options are fine.

xrmx: I'll find out once testing this :)

trentm: > That last part with custom value puts me off a bit.

`db.system` has the same language in semconv:

"`db.system` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used."


So I think using `gen_ai.system` would be a "good thing" (TM).

enrichments/trace/internal/elastic/span.go
```diff
@@ -38,6 +38,12 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
+// Here we are duplicating newer semconv attributes because since v1.26.0
```
codefromthecrypt, line 41: is this comment relevant to this file? specifically the duplicating symbols part

lahsivjar: I would prefer upgrading the semconv to the latest version. If there are any breaking changes in attribute names, we would want to support the updated names, but, even if you don't want to do that I think we can import 2 semconv versions for now. Something like:

```
	semconv25 "go.opentelemetry.io/collector/semconv/v1.25.0"
	semconv26 "go.opentelemetry.io/collector/semconv/v1.26.0"
```

This will make it easier to find which semconv is supported and avoid confusion related to supporting non-semconv fields.

xrmx: @lahsivjar the `NetXXX` attributes we are using here are gone, there are the renamed one `NetworkXXX`. I guess the code should be updated to handle both?

lahsivjar: >  I guess the code should be updated to handle both?

Yes, but that would be outside the scope of this PR. We also don't have a policy on what the support matrix would look like in such cases (like how much older semconv we would need to support).

For now, how about importing both the semconv versions?

enrichments/trace/internal/elastic/span.go
xrmx: Updated to use the gen_ai.system value as span subtype and "genai" as type

enrichments/trace/internal/elastic/span.go
lahsivjar: > https://github.com/elastic/opentelemetry-lib/issues/129

I have created an issue for supporting the newer semconv.

lahsivjar (approved):

enrichments/trace/internal/elastic/span.go
xrmx: @lahsivjar do you want me to already switch the attributes available in 1.27.0 to semconv27 so it's clear what needs to be updated?

lahsivjar: > do you want me to already switch the attributes available in 1.27.0 to semconv27 so it's clear what needs to be updated

I think it would be better as a separate PR.

xrmx:
With the help of @gregkalapos I've setup things to test this but I'm not able to make this work locally.

These are the local changes I have in elastic-collector-components in order to test this:

```
diff --git a/distributions/elastic-components/manifest.yaml b/distributions/elastic-components/manifest.yaml
index b29bf5e..062405c 100644
--- a/distributions/elastic-components/manifest.yaml
+++ b/distributions/elastic-components/manifest.yaml
@@ -10,6 +10,7 @@ extensions:
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/pprofextension v0.114.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage/filestorage v0.114.0
   - gomod: go.opentelemetry.io/collector/extension/memorylimiterextension v0.114.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/basicauthextension v0.114.0
 
 connectors:
   - gomod: github.com/elastic/opentelemetry-collector-components/connector/signaltometricsconnector v0.0.0
diff --git a/processor/elastictraceprocessor/go.mod b/processor/elastictraceprocessor/go.mod
index fc8bd49..29cd501 100644
--- a/processor/elastictraceprocessor/go.mod
+++ b/processor/elastictraceprocessor/go.mod
@@ -19,6 +19,8 @@ require (
        go.uber.org/zap v1.27.0
 )
 
+replace github.com/elastic/opentelemetry-lib => /home/rm/src/opentelemetry-lib
+
 require (
        github.com/cespare/xxhash/v2 v2.3.0 // indirect
        github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
```

Do you see something I am missing?

lahsivjar:
> Do you see something I am missing?

If you are building ocb image then you will need to add replace directive in the ocb config (something like https://github.com/elastic/otel-apm-e2e-poc/blob/f45432edc2613f8997caed7c917ae71f5dec8d4d/ocb.yaml#L32) and then build the custom otel collector to do the testing.

codefromthecrypt:
Curious about the timeline required to get this into the EDOT collector by default once merged? Is it feasible in Jan? (thinking about public demos and/or blogs)

xrmx:
> > Do you see something I am missing?
> 
> If you are building ocb image then you will need to add replace directive in the ocb config (something like https://github.com/elastic/otel-apm-e2e-poc/blob/f45432edc2613f8997caed7c917ae71f5dec8d4d/ocb.yaml#L32) and then build the custom otel collector to do the testing.

Thanks but I get this error. Tried both building v0.114.0 in elastic-collector-components or v0.111.0 in otel-apm-e2e-poc.

```
Error: failed to update go.mod: go subcommand failed with args '[mod tidy -compat=1.22]': exit status 1, error message: go: finding module for package go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/profiles/v1experimental
go: finding module for package go.opentelemetry.io/collector/pdata/internal/data/protogen/profiles/v1experimental
go: go.opentelemetry.io/collector/cmd/builder imports
	go.opentelemetry.io/collector/receiver/otlpreceiver imports
	go.opentelemetry.io/collector/pdata/pprofile/pprofileotlp imports
	go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/profiles/v1experimental: module go.opentelemetry.io/collector/pdata@latest found (v1.21.0), but does not contain package go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/profiles/v1experimental
go: go.opentelemetry.io/collector/cmd/builder imports
	go.opentelemetry.io/collector/connector imports
	go.opentelemetry.io/collector/internal/fanoutconsumer imports
	go.opentelemetry.io/collector/pdata/pprofile imports
	go.opentelemetry.io/collector/pdata/internal/data/protogen/profiles/v1experimental: module go.opentelemetry.io/collector/pdata@latest found (v1.21.0), but does not contain package go.opentelemetry.io/collector/pdata/internal/data/protogen/profiles/v1experimental
```

lahsivjar:
> or v0.111.0 in otel-apm-e2e-poc

Hmm, might be something to do with version up, looks like the otel-e2e-poc has not been updated and otel lib is already using v0.115.0. Let me see if I can do a quick fix there.

xrmx:
:tada: 

Schermata del 2024-12-16 14-42-32

With https://github.com/elastic/otel-apm-e2e-poc/commit/21e7ef1849b6f45e8a0dea788746de6caee0e5ca and by remembering to not start the apm-server works fine. Thanks @lahsivjar and @gregkalapos for the help!

PR #143, opened 2025-01-21
feat: add apm agent configuration package

rogercoll:
Files from https://github.com/elastic/opentelemetry-collector-components/pull/310 so it can be reused in multiple projects.

### DIFF
--- a/agentcfg/elasticsearch.go
+++ b/agentcfg/elasticsearch.go
@@ -56,19 +56,15 @@ const (
 // - Add Otel tracer
 // - Collection metrics
 type ElasticsearchFetcher struct {
-	client        *elasticsearch.Client
-	cacheDuration time.Duration
-
-	mu    sync.RWMutex
-	last  time.Time
-	cache []AgentConfig
-
-	searchSize int
-
+	last             time.Time
+	client           *elasticsearch.Client
+	logger           *zap.Logger
+	cache            []AgentConfig
+	cacheDuration    time.Duration
+	searchSize       int
+	mu               sync.RWMutex
 	invalidESCfg     atomic.Bool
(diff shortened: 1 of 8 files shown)


gregkalapos:
I'd like to raise the discussion of module vs package which we also have in https://github.com/elastic/opentelemetry-lib/pull/141

It seems this lib as a single module gets very big and contains packages with very different concerns. E.g. now it has a new dependency on the elasticsearch client - because that's needed for fetching the config. But all the other packages from this lib don't need that.

Would it make sense to separate things into multiple modules here? And with that, the config fetcher would be a module as well.

rogercoll:
> I'd like to raise the discussion of module vs package which we also have in #141 - [#141 (comment)](https://github.com/elastic/opentelemetry-lib/pull/141#discussion_r1917277748)
> 
> It seems this lib as a single module gets very big and contains packages with very different concerns. E.g. now it has a new dependency on the elasticsearch client - because that's needed for fetching the config. But all the other packages from this lib don't need that.
> 
> Would it make sense to separate things into multiple modules here? And with that, the config fetcher would be a module as well.

I like the idea, I prefer the "extra" complexity of maintaining multiple modules than having to include dependencies that might not be used. But I think we will need to create another issue/pr for this as we will need to adapt the whole repository structure to something similar to https://github.com/elastic/opentelemetry-collector-components/ and release each submodule accordingly.

gregkalapos:
Totally agree on addressing that in a follow up PR - no reason to block this PR for that at this point.

gregkalapos (approved):

rogercoll:
Tracking issue: https://github.com/elastic/opentelemetry-lib/issues/144

PR #149, opened 2025-01-27
feat: add elasticsearch config and client

rogercoll:
- Adds base code for embedding a Go configuration to create Elasticsearch clients. This initial PR contains some of the fields available in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/config.go#L24
- The `configelasticsearch` package follows the same structure as [upstream](https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/confighttp) configuration packages: `NewDefaultConfig` + `ToClient` methods.

Usage (composed type so extra fields can be added):

```go
type ElasticsearchClient struct {
	configelasticsearch.ClientConfig `mapstructure:",squash"`

	CacheDuration time.Duration `mapstructure:"cache_duration"`
}
esClient, err := cfg.Elasticsearch.ToClient(ctx, host, set.TelemetrySettings)
```


config/configelasticsearch/esclient.go
```diff
@@ -0,0 +1,173 @@
...
+		RetryBackoff: createElasticsearchBackoffFunc(&cfg.Retry),
+
+		// configure sniffing
+		DiscoverNodesOnStart:  cfg.Discovery.OnStart,
+		DiscoverNodesInterval: cfg.Discovery.Interval,
+
+		// configure internal metrics reporting and logging
+		EnableMetrics:     false, // TODO
+		EnableDebugLogger: false, // TODO
+		Logger:            &esLogger,
```
axw, line 148: Is the config meant to be backwards compatible with the Elasticsearch exporter upstream?

I've been wondering if this logger should be replaced by https://pkg.go.dev/github.com/elastic/go-elasticsearch/v8#NewOpenTelemetryInstrumentation

rogercoll: > Is the config meant to be backwards compatible with the Elasticsearch exporter upstream?

Correct, all exported fields within the `ClientConfig` structure are a subset of the available ones in the upstream `Elasticsearch exporter` configuration. The idea would be to embed this config in the actual upstream configuration and extent this one as needed.

> I've been wondering if this logger should be replaced by https://pkg.go.dev/github.com/elastic/go-elasticsearch/v8#NewOpenTelemetryInstrumentation

What would be the benefit of replacing the logs for traces (exclusive signal type)? Would it make sense to extend the `TelemetrySettings` to let the user decide on enabling metrics, traces or logs?

At the moment, only logs can be configured:

```yaml
telemetry:
  log_request_body: true
  log_response_body: true
```

I would propose a configuration more aligned with collector's internal telemetry one:

```yaml
telemetry:
  traces:
    enabled: true
  metrics:
    enabled: true
  logs:
    enabled: true
    request_body_enabled: true
    response_body_enabled: true
```

axw: > What would be the benefit of replacing the logs for traces (exclusive signal type)? Would it make sense to extend the TelemetrySettings to let the user decide on enabling metrics, traces or logs?

I think it would be reasonable to add tracing and metrics without taking away logging. There's some overlap with the built-in tracing what those logs are capturing, but it doesn't seem to be a big deal to maintain.

The usual arguments apply for adding tracing, i.e. causality, and being able to attribute time spent in a broader operation/pipeline to Elasticsearch requests. The official Elasticsearch clients have instrumentation, and I believe this is meant to be consistent across them. So you could, for example, create dashboards related to Elasticsearch requests based off traces that works across language clients.

I'm not sure it's necessary to add config for traces/metrics. Could we just rely on the collector's telemetry settings for enabling/disabling?

rogercoll: Thanks for the explanation, I agree on enabling/setting them.

> I'm not sure it's necessary to add config for traces/metrics. Could we just rely on the collector's telemetry settings for enabling/disabling?

Good point, do you know if there is a way to check if collector's telemetry is enabled? It looks like the `TelemetrySettings.TracerProvider` is always non-nil, even if `telemetry::traces::level::none` → `nopTracer`. 

This commit https://github.com/elastic/opentelemetry-lib/pull/149/commits/f0ba3711598fca7973cf96b04b5d7954f18b8e7e sets the internal tracing. By default, it won't capture the search body. I would suggest adding the exported telemetry configuration options (and metrics) as we need them or in a separate PR, that way we can have a wider agreement on the naming.

axw: > Good point, do you know if there is a way to check if collector's telemetry is enabled? It looks like the TelemetrySettings.TracerProvider is always non-nil, even if telemetry::traces::level::none → nopTracer.

I'm not aware of a way to do this. Could you do an equality comparison to `noop.NewTracerProvider()`? Otherwose I think the best thing would be for the instrumentation to check if the span is recording before doing anything expensive -- probably best for it to do that anyway, so it's inexpensive when using head-based sampling.

> This commit https://github.com/elastic/opentelemetry-lib/commit/f0ba3711598fca7973cf96b04b5d7954f18b8e7e sets the internal tracing. By default, it won't capture the search body. I would suggest adding the exported telemetry configuration options (and metrics) as we need them or in a separate PR, that way we can have a wider agreement on the naming.

++ sounds good

config/configelasticsearch/esclient.go
```diff
@@ -147,10 +149,31 @@ func (cfg *ClientConfig) ToClient(
...
+		return true
+	}
+
+	// The actual implementation of the no-op tracer (tracer.noopNoContextTracer)
+	// is not exported by the OpenTelemetry collector, so we cannot directly check for it.
+	// Instead, we create a span and verify whether it records data.
+	//
+	// If the tracer is a no-op, the span will not record any events or attributes.
+	// We avoid ending the span to prevent unnecessary delivery.
+	_, span := provider.Tracer("es-config").Start(context.Background(), "noop-test")
```
rogercoll, line 173: @dmathieu Are you aware of any mechanism to verify if the tracer provider is NO-OP without leaking this span?

cc @gregkalapos

dmathieu: Another side effect of this is that if sampling is enabled, the span may not be recording, the tracer is enabled though.

dmathieu: There's no way to retrieve this information cleanly. If you really want to do this, you could possibly print a `%#v` of the tracer, and check that it is a `&trace.tracer{`.
That won't work if folks are using a custom SDK though (such as [apmotel](https://github.com/elastic/apm-agent-go/tree/main/module/apmotel)).

In the future, this can be checked with the in-development [Enabled](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#enabled) method.
That specification is not yet enabled in the Go SDK though.

rogercoll: Good point, maybe another option would be checking the span type which seems to be exported (`"go.opentelemetry.io/otel/trace/noop"`)

```suggestion
	_, span := provider.Tracer("es-config").Start(context.Background(), "noop-test")
	_, ok := span.(noop.Span)
	return ok

```

dmathieu: That still won't work with custom SDKs.

dmathieu: That could work, yes.

For the record, if a span is not sampled, a non-recording span is returned, but that's not a noop.
https://github.com/open-telemetry/opentelemetry-go/blob/007fb1f65bdd5df6e0ea851c7b05ca047bc48f4e/sdk/trace/tracer.go#L152

rogercoll: > There's no way to retrieve this information cleanly. If you really want to do this, you could possibly print a %#v of the tracer, and check that it is a &trace.tracer{.
That won't work if folks are using a custom SDK though (such as [apmotel](https://github.com/elastic/apm-agent-go/tree/main/module/apmotel)).

I am not aware of the collector allowing custom SDKs through its configuration, and it looks like that if telemetry is not disabled it uses the Go SDK tracer provider which is an exported type https://github.com/open-telemetry/opentelemetry-collector/blob/v0.118.0/service/telemetry/tracer.go#L66 

Since the `Enabled()` option is not available yet, we could enable the instrumentation only if the sdk tracer is used: 
https://github.com/elastic/opentelemetry-lib/pull/149/commits/2b88a3dd90543159f95bfed79744428a0d6df3a7

carsonip (approved):
code lgtm. Are you planning to make es exporter use this config and client?

rogercoll:
> code lgtm. Are you planning to make es exporter use this config and client?

The initial plan is to use it for a couple of extensions in Elastic's components repository, but yes, I will create an issue upstream to follow up on the exporter too.

Issue #169, opened 2025-03-13
Handle characters in service.name that'd be disallowed in APM Server


gregkalapos:
In the classic Elastic APM Data model the [allowed characters in `service.name` are](https://github.com/elastic/apm-agent-nodejs/blob/main/docs/configuration.asciidoc): `a-z, A-Z, 0-9, -, _, and space.`

OTel allows basically any printable characters (this part is TBD - we need to look into what is exactly allowed). 

The issue with this is that once we ingest characters that are not allowed in the classic APM Data model, Kibana may break. One specific example found by @intuibase is `/`.

Goal of the issue: Implement a logic in EDOT - (so far the idea is to do it specifically in the enrichment in this repo) to normalize these characters.

smith:
We have an open issue to be able to handle these characters in Kibana: https://github.com/elastic/kibana/issues/213943

> Goal of the issue: Implement a logic in EDOT - (so far the idea is to do it specifically in the enrichment in this repo) to normalize these characters.

Will normalization be necessary if the UI can handle these characters?

gregkalapos:
https://github.com/elastic/kibana/issues/213943 and this one are somewhat competing solutions for the problem.

> Will normalization be necessary if the UI can handle these characters?

No, it won't be needed if Kibana can deal with this.

Reason I opened this because we can open an issue for this fairly quickly and based on what @intuibase says this seems like a fairly big limitation for the PHP OTel SDK. 


Overall I think https://github.com/elastic/kibana/issues/213943 is a much nicer solution - but then the question is how long would that take.

Other thing to consider is that if we solve this in EDOT, then all older Kibana versions will also work, while if we go with https://github.com/elastic/kibana/issues/213943, then only Kibana versions with PR addressing this issue won't have the problem.

There was a slack thread about this, I'll tag you there @smith.

gregkalapos:
Opened https://github.com/elastic/opentelemetry-lib/pull/170 - that's how it would look like if this is fixed in EDOT. The logic is the same as it was with APM Server.

If UI folks say they'd prefer going with https://github.com/elastic/kibana/issues/213943, I'd definitely support that and I'd be happy to close https://github.com/elastic/opentelemetry-lib/pull/170 and this issue without merging - since solving this in Kibana and not changing values coming in via OTLP is a nicer solution.

Let me know how we should proceed.

gregkalapos:
Closing in favour of https://github.com/elastic/kibana/issues/213943.

PR #170, opened 2025-03-14
Normalize service.name as apm-data does

gregkalapos:
Solves https://github.com/elastic/opentelemetry-lib/issues/169.

Before merge, we need to decide if we go with this solution or we prefer https://github.com/elastic/kibana/issues/213943 and don't need this. I'll wait with merging until we have clear agreement.

This PR normalizes `service.name` the same way [as `apm-data` does](https://github.com/elastic/apm-data/blob/main/input/otlp/metadata.go#L489). 

This is especially relevant for values where `service.name` would expect values that Kibana can't handle currently - such example is `/` which seems to be used by the PHP OTel SDK.


gregkalapos:
Closing in favour of https://github.com/elastic/kibana/issues/213943.

Issue #172, opened 2025-03-19
[enrichments] Add enrichment logic for `user_agent.*`


lahsivjar:
These sets of attributes are not always set by the OTel SDKs but are required by the UI. The below OTTL statements can be used as an algorithm for populating these fields based on stable OTel semconv:

```
transform/user_agent:
    trace_statements:
      - merge_maps(span.cache, UserAgent(span.attributes["user_agent.original"]), "upsert") where span.attributes["user_agent.original"] != nil
      - set(span.attributes["user_agent.name"], span.cache["user_agent.name"])
      - set(span.attributes["user_agent.version"], span.cache["user_agent.version"])
      - set(span.attributes["url.full"], Concat([span.attributes["url.scheme"], "://", span.attributes["server.address"]], ""))
      - set(span.attributes["url.full"], Concat([span.attributes["url.full"], span.attributes["server.port"]], ":")) where span.attributes["server.port"] != nil
      - set(span.attributes["url.full"], Concat([span.attributes["url.full"], span.attributes["url.path"]], "")) where span.attributes["url.path"] != nil
      - set(span.attributes["url.full"], Concat([span.attributes["url.full"], span.attributes["url.query"]], "?")) where span.attributes["url.query"] != nil
```

In the above OTTL statements, the [UserAgent converter func](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/func_useragent.go#L9C3-L9C39) uses https://github.com/ua-parser/uap-go which can be used by this library to do the same enrichment.

Based on recent developments, the `url.full` will be built in the UI so we only need to do the user_agent bit in the opentelemetry lib.

PR #174, opened 2025-03-21
Add enrichment for user_agent.{name, version}

lahsivjar:
Adds enrichment for `user_agent.{name, version}` if it is missing. The attributes are derived from `user_agent.original`.

Closes: https://github.com/elastic/opentelemetry-lib/issues/172

### DIFF
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/elastic/opentelemetry-lib/elasticattr"
 	"github.com/elastic/opentelemetry-lib/enrichments/trace/config"
+	"github.com/ua-parser/uap-go/uaparser"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	semconv25 "go.opentelemetry.io/collector/semconv/v1.25.0"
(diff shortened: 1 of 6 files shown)


enrichments/trace/internal/elastic/span.go
```diff
@@ -520,6 +553,15 @@ func (s *spanEnrichmentContext) setInferredSpans(span ptrace.Span) {
 	}
 }
 
+func (s *spanEnrichmentContext) setUserAgentIfRequired(span ptrace.Span) {
+	if s.userAgentName == "" && s.inferredUserAgentName != "" {
+		span.Attributes().PutStr(semconv27.AttributeUserAgentName, s.inferredUserAgentName)
+	}
+	if s.userAgentVersion == "" && s.inferredUserAgentVersion != "" {
+		span.Attributes().PutStr(semconv27.AttributeUserAgentVersion, s.inferredUserAgentVersion)
+	}
```
lahsivjar, line 562: [**For reviewers**] I am setting the `user_agent.{name, version}` only if it is not already set.

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -364,6 +365,63 @@ func TestElasticTransactionEnrich(t *testing.T) {
 				return &spanLinks
 			}(),
 		},
+		{
+			name: "user_agent_parse_name_version",
```
gregkalapos, line 369: nit: I'd postfix these with something like `_span` and `_transaction`. So e.g. `user_agent_parse_name_version_transaction` here and `user_agent_parse_name_version_span` below, instead of having them with the same `name`.

gregkalapos (approved):
lgtm

enrichments/trace/internal/elastic/span_test.go
lahsivjar: They would be reported as `TestElastic{Transaction, Span}Enrich/user_agent_parse_name_version`, is that not enough or is it for when looking at the tests individually without the context of the main test?

lahsivjar: Cool, I am thinking to simplify these tests with a method for default enrichments... currently the tests are getting a bit hard to read an necessarily verbose. Will do it later as a followup.

PR #180, opened 2025-04-16
Fix handling of non root span that represent a dependency

gregkalapos:
Regression reported by @AlexanderWert.

Problem was missing `(` `)` when dependencies were calculated resulting in mapping a non-root span partly as a transaction and not setting `processor.event`.

Also added a test covering such cases.

### DIFF
--- a/enrichments/trace/internal/elastic/span.go
+++ b/enrichments/trace/internal/elastic/span.go
@@ -220,7 +220,7 @@ func (s *spanEnrichmentContext) enrich(span ptrace.Span, cfg config.Config) {
 	// In OTel, a local root span can represent an outgoing call or a producer span.
 	// In such cases, the span is still mapped into a transaction, but enriched
 	// with additional attributes that are specific to the outgoing call or producer span.
-	isExitRootSpan := s.isTransaction && span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer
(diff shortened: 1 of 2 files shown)


gregkalapos:
Since this causes missing spans (as `processor.event` is not set), we should make sure this is included in `9.0.1` und `8.18.1`. 

@lahsivjar assuming the PR is ok, could we push a release of this repo in the near future?

enrichments/trace/internal/elastic/span.go
```diff
@@ -220,7 +220,7 @@ func (s *spanEnrichmentContext) enrich(span ptrace.Span, cfg config.Config) {
 	// In OTel, a local root span can represent an outgoing call or a producer span.
 	// In such cases, the span is still mapped into a transaction, but enriched
 	// with additional attributes that are specific to the outgoing call or producer span.
-	isExitRootSpan := s.isTransaction && span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer
+	isExitRootSpan := s.isTransaction && (span.Kind() == ptrace.SpanKindClient || span.Kind() == ptrace.SpanKindProducer)
```
lahsivjar, line 223: 🤦 thanks a lot for finding and fixing this.

lahsivjar (approved):

lahsivjar:
> @lahsivjar assuming the PR is ok, could we push a release of this repo in the near future?

I can cut a new release, the dependabot should do the rest.

PR #184, opened 2025-04-24
Mobile transaction type

LikeTheSalad:
Fixes https://github.com/elastic/kibana/issues/216980

These changes aim to set `transaction.type=mobile` to all root spans coming from our mobile agents.


LikeTheSalad:
### Question

These changes set `transaction.type` to `mobile` based on the [elastic-specific type](https://github.com/elastic/apm/tree/main/specs/agents/mobile#additional-span-attributes) attribute. Should we remove the `type` attribute from the span after we use it to set `transaction.type`?

gregkalapos (approved):
lgtm. Thanks for the PR! 🎉 

>These changes set `transaction.type` to `mobile` based on the [elastic-specific type](https://github.com/elastic/apm/tree/main/specs/agents/mobile#additional-span-attributes) attribute. Should we remove the `type` attribute from the span after we use it to set `transaction.type`?

I'd keep it. In this lib, we typically don't remove any attributes. We could save some storage if we remove it, on the other hand, it'd be very confusing if we start removing fields from incoming data. So my take is to keep it.

enrichments/trace/internal/elastic/span.go
```diff
@@ -200,6 +201,10 @@ func (s *spanEnrichmentContext) Enrich(
 			s.userAgentName = v.Str()
 		case semconv27.AttributeUserAgentVersion:
 			s.userAgentVersion = v.Str()
+		case "type":
```
lahsivjar, line 204: How will this attribute be set?

gregkalapos: It's set by the EDOT mobile SDKs [specified here](https://github.com/elastic/apm/blob/main/specs/agents/mobile/README.md#additional-span-attributes).

enrichments/trace/internal/elastic/span.go
```diff
@@ -200,6 +201,10 @@ func (s *spanEnrichmentContext) Enrich(
...
+			if v.Str() == "mobile" {
```
gregkalapos, line 205: @lahsivjar pointed out on Slack that APM Server sets `transaction.type` always, when attribute `type` is present to that value. Seems to be done here: https://github.com/elastic/apm-data/blob/main/input/otlp/traces.go#L452-L453

I understand we want to solve the `transaction.type` issue for mobile. But to be aligned with APM Server, I'd propose we do the same as apm-data: always take the value from the attribute `type` and set that to `transaction.type` - not only for the value `mobile`.

LikeTheSalad: I've just applied this suggestion, cheers!

enrichments/trace/internal/elastic/span.go
felixbarny: Maybe also add `elasticattr.TransactionType` do this case so that if `transaction.type` is set already, it's not overridden?

enrichments/trace/internal/elastic/span.go
```diff
@@ -319,7 +324,7 @@ func (s *spanEnrichmentContext) enrichSpan(
 		s.setUserAgentIfRequired(span)
 	}
 
-	if isExitRootSpan && transactionTypeEnabled {
+	if isExitRootSpan && transactionTypeEnabled && !s.hasType {
```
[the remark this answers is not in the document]
LikeTheSalad: The logic here causes [this issue](https://github.com/elastic/kibana/issues/216980). Even if the `type` is set, the logic in the if body overwrites it. Making [this test](https://github.com/elastic/opentelemetry-lib/pull/184/files#diff-d3200800e648faa320aec93f9d104ec7daf33e8d5f36a14aed5ebf0532dd6407R524) to fail.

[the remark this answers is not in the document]
lahsivjar: Hmm, this case should then affect all the transactions which are classified as exit root spans. So, every exit root span will override the transaction type set by the span type - I can't think of a reason why we should have the logic differ if the type is explicitly provided. I am not sure about the specs around the exit root span but that doesn't feel right to me @gregkalapos Can you shed some light on this? Do we need to have a check here to skip overriding the transaction type for exit root spans if they are already set or should we always be overriding?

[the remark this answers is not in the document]
gregkalapos: The thing with exit root spans is that they typically won't match our default logic to derive `transaction.type`. So without this logic here, these'd either end up with `unknown` or worse with a confusing `transaction.type` value.

Examples:
- Root database call -> `transaction.type` would be `unknown`
- Root outgoing HTTP call -> would end up with `transaction.type` `request`, this is an even more confusing one, because we set it to the same value in backend agents for incoming HTTP calls - so those traces starting with an outgoing HTTP call would end up in the group with backend incoming transactions.

Hence the idea to derive `transaction.type` for exit root spans based on span type and subtype. For the examples above with this logic `transaction.type` will be `db.elasticsearch` and `external.http`.

But this breaks for mobile, because - Kibana mobile UI expects `transaction.type` `mobile`, otherwise those traces won't even show up.

Prior [to my proposal](https://github.com/elastic/opentelemetry-lib/pull/184#discussion_r2059752323) we only had this exception for mobile. But now, as you say @lahsivjar we always do this, when the type is explicitly provided. But thinking about it, I'd say, it's still good.

With this, we'd still apply the logic described above to have more meaningful `transaction.type` for exit root spans - except when the type is explicitly provided - in that case, we don't change the value, but keep it as is.

So overall, this seems ok to me.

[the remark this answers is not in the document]
lahsivjar: Thanks for the explanation - it sounds good to me now!

enrichments/trace/internal/elastic/span.go
```diff
@@ -98,6 +99,7 @@ type spanEnrichmentContext struct {
 	isDB                     bool
 	messagingDestinationTemp bool
 	isGenAi                  bool
+	hasType                  bool
```
lahsivjar, line 102:
```suggestion
```

enrichments/trace/internal/elastic/span.go
```diff
@@ -319,7 +323,7 @@ func (s *spanEnrichmentContext) enrichSpan(
 		s.setUserAgentIfRequired(span)
 	}
 
-	if isExitRootSpan && transactionTypeEnabled {
+	if isExitRootSpan && transactionTypeEnabled && !s.hasType {
```
lahsivjar, line 326:
```suggestion
	if isExitRootSpan && transactionTypeEnabled && s.typeValue == "" {
```

lahsivjar (approved):
Missed a couple of places to remove `hasValue` variable. Rest LGTM!

enrichments/trace/internal/elastic/span_test.go
```diff
@@ -446,6 +446,138 @@ func TestElasticTransactionEnrich(t *testing.T) {
 	}
 }
 
+// Tests the enrichment logic for elastic's transaction definition with type.
+func TestElasticTransactionEnrichWithType(t *testing.T) {
```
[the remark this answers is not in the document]
LikeTheSalad: Got it. I've just moved the test cases to existing tests.

enrichments/trace/internal/elastic/span.go
LikeTheSalad: > Maybe also add `elasticattr.TransactionType` do this case so that if `transaction.type` is set already, it's not overridden?

I'm ok with validating that `transaction.type` won't change if it's already present. It's a different use case though it seems like it'd be easy enough to add it in this PR. What do you think, @gregkalapos @lahsivjar ?

lahsivjar: > Maybe also add elasticattr.TransactionType do this case so that if transaction.type is set already, it's not overridden?

I think overriding behaviour should be handled for other attributes too, but I am not sure if not overriding in all cases would be the correct option. I have created https://github.com/elastic/opentelemetry-lib/issues/185 to discuss it further.

lahsivjar (approved):
LGTM!

enrichments/trace/internal/elastic/span.go
gregkalapos: Ok by me to include it in this PR.

LikeTheSalad: > I think overriding behaviour should be handled for other attributes too, but I am not sure if not overriding in all cases would be the correct option. I have created https://github.com/elastic/opentelemetry-lib/issues/185 to discuss it further.

Got it. Thanks, @lahsivjar. I won't add that change here then.

lahsivjar: I think we should do it in a separate PR since this concern might apply to other attributes too, but I don't have a strong opinion here. Does that makes sense @gregkalapos @felixbarny ? Or do we think only `transaction.type` needs to be handled this way?

PR #204, opened 2025-07-08
enrichments: handle gen_ai.provider.name too

xrmx:
>[!IMPORTANT]
This is in draft because semconv 1.37.0 has not been released yet and then also not bumped in the dependencies 

Same enrichment as the previous gen_ai.system


lahsivjar (approved):
LGTM! I don't have full context on gen AI instrumentation, and, while it seems like a very straightforward PR if you do require any technical review on the usage of the field please wait for other reviews.

